PR #4, opened 2021-04-08
Allow package to be installed

jayvdb:


setup.py
```diff
@@ -35,4 +34,7 @@
         "Topic :: Utilities",
     ],
     test_suite="tests",
+    entry_points={
```
adsharma, line 37: I find a locally runnable python script convenient for testing before installation.

The rename to cli.py is fine, but could you add back the lines removed, so that part works the same way as before?

adsharma (commented):
Thanks for taking an interest in the project. Yes, it would be nice to make it installable and accessible via pypi

setup.py
jayvdb: Sure can

jayvdb: oh, you mean in ./py2many.py ; let me see if something similar will suffice.

jayvdb: you can always do `PYTHONPATH={PWD} python3 ./py2many.py` , but I have added some glue so `PYTHONPATH` doesnt need to be set.

jayvdb: ping @adsharma , how does this look now?

adsharma: Looks great. Thank you!

PR #9, opened 2021-04-09
Re-add directory mode 

jayvdb:


py2many/cli.py
```diff
@@ -131,11 +140,40 @@ def main():
...
             outdir = pathlib.Path(args.outdir)
-        print(f"Writing to: {outdir}")
-        output_path = outdir / (source.stem + settings.ext)
-        print(f"{filename}...{output_path}")
-        with open(output_path, "w") as f:
-            source_data = open(source).read()
-            f.write(transpile(source_data, settings.transpiler))
-        os.system(f"{settings.formatter} {output_path}")
-        print()
+        if os.path.isfile(filename):
```
adsharma, line 143:
```
source.is_file()
```

Some of the code below reads a lot nicer with pathlib:

https://realpython.com/python-pathlib/

py2many/cli.py
```diff
@@ -131,11 +140,40 @@ def main():
...
+            print(f"Writing to: {outdir}")
+            _process_once(settings, source, outdir)
+        else:
+            basename = os.path.basename(filename)
+            if args.outdir is None:
+                outdir = source.parent / f"{basename}-py2many"
```
adsharma, line 149:
```
 outdir = source.parent / f"{source.name}-py2many"
 ```

File: py2many/cli.py [hunk 3]
```diff
@@ -131,11 +140,40 @@ def main():
...
+            print(f"Transpiling whole directiory to {outdir}:")
+            successful = failures = format_errors = 0
+            for root, subdir, filenames in os.walk(filename):
```
adsharma, line 152: Could be written is using source.glob("*.py")

py2many/cli.py
```diff
@@ -90,6 +90,15 @@ def go_settings(args):
     return LanguageSettings(GoTranspiler(), ".go", "gofmt -w")
 
 
+def _process_once(settings, filename, outdir):
+    output_path = outdir / (filename.stem + settings.ext)
+    print(f"{filename}...{output_path}")
+    with open(output_path, "w") as f:
+        source_data = open(filename).read()
```
adsharma, line 97: File handle leak here (leak present in main branch as well). Using a context manager should fix.

adsharma (changes requested):
Would love to have the directory mode.

py2many/cli.py

PR #10, opened 2021-04-10
Add (Enum, str) base class support

jayvdb:


pyrs/transpiler.py
```diff
@@ -333,6 +333,10 @@ def visit_IntEnum(self, node):
         fields = "\n".join(fields)
         return f"enum {node.name} {{\n{fields}\n}}\n\n"
 
+    def visit_StrEnum(self, node):
+        self._usings.add("strum_macros::EnumString")
+        return f"#[derive(Debug, PartialEq, EnumString)]\n{self.visit_IntEnum(node)}"
```
adsharma, line 338: Rename logic to `visit_Enum` or `visit_CommonEnum` and then have Int/StrEnum reuse the same code.
Also a test case similar to int_enum.py for code coverage in the same commit.

adsharma (changes requested):
Typo in the description: Emum -> Enum

adsharma:
Merged thanks!

jayvdb:
Actually you merged your own version of this, and the generated code failed to compile.

PR #11, opened 2021-04-10
Improve Rust and Nim module level constant support

jayvdb:
Also improve the Vec declaration.

Sets on top of #81


jayvdb:
@adsharma i've marked this ready for review now.

tests/expected/binit.rs
```diff
@@ -16,10 +16,10 @@ fn bisect_right(data: Vec<i32>, item: i32) -> i32 {
 
 fn bin_it(limits: Vec<i32>, data: Vec<i32>) -> Vec<i32> {
     let mut bins = vec![0];
-    for _x in limits {
+    for _x in limits.iter() {
```
adsharma, line 19: This is equivalent to:

```
for _x in limits
```
if limits is &Vec<i32> which should be the case to match python's reference semantics. I have a diff in flight to make it happen.


or

```
for _x in &limits  # if limits is a Vec<T>
```

The conversion to `iter()` could be implicit.

adsharma (changes requested):
Neat improvement!

Summary:

* Let's deal with cleaning up of type maps in a separate diff
* nim var vs let vs const for globals needs more analysis of the python code. Let's drop the "global variables are const" for now
* Different fix to binit.rs in flight.

tests/expected/binit.rs
jayvdb: This is needed for arrays.  Otherwise
```rust
error[E0277]: `[i32; 2]` is not an iterator
 --> global.rs:8:14
  |
8 |     for i in l_a {
  |              ^^^ borrow the array with `&` or call `.iter()` on it to iterate over it
  |
  = help: the trait `Iterator` is not implemented for `[i32; 2]`
  = note: arrays are not iterators, but slices like the following are: `&[1, 2, 3]`
  = note: required because of the requirements on the impl of `IntoIterator` for `[i32; 2]`
  = note: required by `into_iter`

error[E0277]: `[&str; 2]` is not an iterator
  --> global.rs:11:14
   |
11 |     for i in l_b {
   |              ^^^ borrow the array with `&` or call `.iter()` on it to iterate over it
   |
   = help: the trait `Iterator` is not implemented for `[&str; 2]`
   = note: arrays are not iterators, but slices like the following are: `&[1, 2, 3]`
   = note: required because of the requirements on the impl of `IntoIterator` for `[&str; 2]`
   = note: required by `into_iter`
```

Anyway, I will split off the Vec -> array stuff to a separate PR.

jayvdb: grrr.  I can not.  `Vec` can not be a global in Rust.

```rust
3 | const l_a = vec![code_0, code_1];
  |       ^^^ help: provide a type for the item: `l_a: <type>`

error: missing type for `const` item
 --> /Users/john.vandenberg/python/py2many/tests/cases/global.rs:6:7
```

```rust
error: missing type for `static` item
 --> /Users/john.vandenberg/python/py2many/tests/cases/global.rs:3:8
  |
3 | static l_a = vec![code_0, code_1];
  |        ^^^ help: provide a type for the item: `l_a: <type>`
```

pynim/transpiler.py
```diff
@@ -495,11 +503,8 @@ def visit_Assign(self, node):
             target = self.visit(target)
             value = self.visit(node.value)
 
-            if len(node.scopes) == 1:
-                if isinstance(
-                    node.scopes[0], ast.Module
-                ):  # if assignment is module level it must be const
-                    return f"const {target} = {value};"
+            if is_global(node):  # if assignment is module level it must be const
```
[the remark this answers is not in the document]
jayvdb: Yup, agreed.

PR #12, opened 2021-04-10
Expand cli test cases

jayvdb:


adsharma:
This looks great. CI is much needed and your changes to test_cases.py also look reasonable (assuming all tests work the same way as before).

* When the formatter is made optional, how do we know that the generated code is valid? Right now we rely on the formatter so the comparison to golden output succeeds.

* The code needs rebasing so I can merge.

jayvdb:
The formatter is Optional, but not really optional ;-).  The primary goal of that commit is to replace "/bin/true" which is a Linux-ism that isnt appropriate for Windows or macos.

We would still use and require a formatter where one exists.

tests/test_cli.py
```diff
@@ -34,9 +34,15 @@ def test_cli(self, case, lang):
...
-            with open(f"expected/{case}{ext}") as f2:
-                with open(f"cases/{case}{ext}") as actual:
-                    self.assertEqual(f2.read(), actual.read())
+            with open(f"cases/{case}{ext}") as actual:
+                generated = actual.read()
+                if os.path.exists(f"expected/{case}{ext}"):
+                    with open(f"expected/{case}{ext}") as f2:
+                        self.assertEqual(f2.read(), generated)
+
+                if not os.path.exists(f"expected/{case}{ext}"):
```
adsharma, line 43: Yes - this would be useful functionality. But I'd make it an explicit action (add a new flag for example), rather than checking for the presence of a file.

jayvdb: This only occurs in the scenario were someone is adding new test cases.  That is an explicit action that person is taking.

I've expanded the concept further at https://github.com/adsharma/py2many/pull/29/files#diff-4e8715c7a425ee52e74b7df4d34efd32e8c92f3e60bd51bc2e1ad5943b82032eR86

And we can continue to improve it as the test suite and test rig develops.  Please get this in, and activate GitHub actions, as I have lots of pending changes not yet submitted and it is annoying to be carrying so many changes around in so many branches.

adsharma: Perfect thanks! 

The only suggestion I have is to make those two env variables into command line flags, where the default could still come from env.

Also the commit needs rebasing.

jayvdb: I am not going to make them command line flags.  You can do that yourself, later.

Please be more considerate of other peoples time and effort helping your project.  You keep causing conflicts with your own commits, that mean I need to rebase my PRs, which is an especially large time waster when I have many PRs stacked on top of each other.

jayvdb: OMG.  You have broken my PR again!

Please stop merging your own commits ahead of this!

adsharma:
Thanks for rebasing. Not trying to create any more work for you than needed.

It's not usual for such a thing to happen in an open source project with more than one person committing code.

PR #22, opened 2021-04-11
MANIFEST.in: Create

jayvdb:


adsharma:
Could you point me to some documentation on why this is necessary?

https://www.remarkablyrestrained.com/python-setuptools-manifest-in/

thinks setup.py can do the same thing.

jayvdb:
Yes, you can possibly use setuptools_scm instead.  That has other benefits, but it also has negatives, such as requiring setuptools_scm is installed before setup.py can be run.

PR #35, opened 2021-04-11
Various print fixes

jayvdb:


jayvdb:
These fixes appear to have been cherry-picked onto the main branch.

PR #29, opened 2021-04-11
tests: Run compile on generated files 

jayvdb:
Run compile on generated files.

Sits on top of https://github.com/adsharma/py2many/pull/12

PR #38, opened 2021-04-12
Remove unused f-strings and imports

jayvdb:
Replace f-strings with normal string, remove star imports and remove unused imports.

Sits on top of https://github.com/adsharma/py2many/pull/29


jayvdb:
If you fast-forward this, #29 will also be merged.

Issue #30, opened 2021-04-11
Duplicate main


jayvdb:
The follow results in two `main()` functions
```py

def main():
	...


if __name__ == "__main__":
	main()
```

Generated rust:
```rust

fn main() {
    ...
}

fn main() {
    main();
}
```

adsharma:
I think the fix for this is to rename all other instances of "main" in the module namespace.
One way to implement this is to write a new transformer which checks for the presence of:

```
if __name__ == "__main__":
```

and if present, do a renaming pass over the entire AST.

Issue #25, opened 2021-04-11
Python test cases should be runnable scripts


jayvdb:
In addition to being compile-able, the generated files should also be runnable and their output should be the same as the Python script.  That adds an extra layer of validation.

jayvdb:
https://github.com/adsharma/py2many/pull/35 is needed to get print working on each target, so we can compare them to the python baseline.

jayvdb:
There are still quite considerable differences in how the print function transpiles. Might be better getting `assert` working - should be easier to make it consistent when the assertion is true.

jayvdb:
`assert` isnt supported by many of the target languages.

So it seems the quickest path to getting invocation happening is to inject "empty" `main()` , and maybe fix the bugs related to `pass` so that can be used to make the `main()` truly empty in a consistent way.

jayvdb:
When the current set of PRs are merged, all tests are runnable & comparable and are run in CI on at least one platform.

The `is_script` part of the test rig is then redundant, and installing rust-script is no longer necessary.

The test rig can then be rewritten to make that mandatory for all new test cases, making `INVOKER` the "primary" focus of the test logic (and `COMPILER` an optional step), and maybe even enhanced to require that each test case at least passes for at least one language.

jayvdb:
@adsharma , please review https://github.com/adsharma/py2many/pull/60 and https://github.com/adsharma/py2many/pull/85 (and would you mind restarting that job; the failure looks very unrelated to the changes in the PR.)

jayvdb:
I guess one merge per day is all I am allowed :P

adsharma:
They're all runnable scripts now.

PR #47, opened 2021-04-13
ci: Invoke generated code

jayvdb:


jayvdb:
@adsharma this should be ready for review now

tests/test_cli.py
```diff
@@ -13,15 +13,29 @@
...
 UPDATE_EXPECTED = os.environ.get("UPDATE_EXPECTED", False)
 COMPILERS = {
-    # cpp is disabled due to https://github.com/adsharma/py2many/issues/24
-    # "cpp": ["clang", "-std=c++14"],
-    "dart": ["dart"],
+    "cpp": ["clang", "-std=c++14"],
+    "dart": ["dart", "compile", "exe"],
     "go": ["go", "tool", "compile"],
-    "julia": ["julia"],
-    "kotlin": ["kotlinc", "-p", "library"],
```
adsharma, line 21: "-p library" allows kotlin to be compiled without main()

tests/test_cli.py
```diff
@@ -40,22 +54,25 @@
     "lambda.nim",  # https://github.com/adsharma/py2many/issues/27
     "lambda.rs",  # https://github.com/adsharma/py2many/issues/15
     "str_enum.jl",  # https://github.com/adsharma/py2many/issues/26
-]
+] + [f"{case}.cpp" for case in TEST_CASES]
+# See https://github.com/adsharma/py2many/issues/24 for cpp failures
+
+
+def has_main(filename):
```
adsharma, line 61: Yeah - this is simple and effective. Good for now.

In the long run, handling this via a rename pass would avoid any issues caused by dealing with strings vs AST

.github/workflows/main.yml
```diff
@@ -13,16 +13,47 @@ on:
...
+          elif [ "$RUNNER_OS" == "macOS" ]; then
+              brew tap holgerbrandl/tap https://github.com/holgerbrandl/homebrew-tap
+              brew install kscript ktlint
+          elif [ "$RUNNER_OS" == "Windows" ]; then
+              choco install zip curl
+          fi
+          dart pub global activate stagehand
+          curl -s "https://get.sdkman.io" | bash
+          source "$HOME/.sdkman/bin/sdkman-init.sh"
+          sdk install kscript
```
adsharma, line 48: Does this requires a macOS check? It seems to be installed via brew a few lines above.

adsharma (commented):
Overall effect of this PR:

```
before:
================= 110 passed, 37 skipped, 5 warnings in 23.30s =================
after:
================= 89 passed, 58 skipped, 5 warnings in 45.09s ==================
```

My reading is that the tests got more stringent - it's not enough to generate the code and run the formatter. We're now trying to compile and invoke the code.

For the 110-89 = 21 tests, could you confirm that any future regressions will be reported as failures instead of "skipped"?

All the comments I have are relatively minor and could be addressed in subsequent diffs if any change is needed.

tests/test_cli.py
jayvdb: `kotlinc-jvm` doesnt support that.

tests/test_cli.py
jayvdb: In the long run we do https://github.com/adsharma/py2many/issues/25 and https://github.com/adsharma/py2many/issues/30, and this `has_main` gets removed.

.github/workflows/main.yml
jayvdb: macOS CI is a few seconds slower because of the duplication.  Windows is the one that is slow.

the "install OS deps" should be split into 3 per-OS tasks that use `if` yaml guards to ensure they run on only one OS.

PR #48, opened 2021-04-13
Use implicit super()

jayvdb:
Explicit super are unnecessary when writing Python 3+ only
code.  Removing them means the reader doesn't need to check
if the super args are doing anything funky.

### DIFF
--- a/py14/transpiler.py
+++ b/py14/transpiler.py
@@ -209,13 +209,13 @@ class CppTranspiler(CLikeTranspiler):
 
     def visit_Str(self, node):
         """Use a C++ 14 string literal instead of raw string"""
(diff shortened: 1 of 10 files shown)


jayvdb:
was merged into main as part of another PR which wasnt ff merged, so the commits were different and this PR was left dangling.

adsharma:
This one needed a rebase because I had pushed https://github.com/adsharma/py2many/commit/8168d9a451a348256cbe2a339215feac2a402dbd to fix the pyflakes error before I saw the PR.

Also an explanation for why github green button options are all bad:

https://stackoverflow.com/questions/60597400/how-to-do-a-fast-forward-merge-on-github

PR #49, opened 2021-04-13
ci: Add black & pyflakes

jayvdb:
Sits on top of https://github.com/adsharma/py2many/pull/48
but that is optional - I can rebase onto main if pr 48
is not desirable or needs adjustments.


jayvdb:
rebased. ping @adsharma

PR #52, opened 2021-04-14
Fixed width ints

adsharma:


jayvdb (approved):
could you add a `def main` and run `kscript...` from https://github.com/adsharma/py2many/pull/47 to ensure it is actually working. if so, lets get this in before that PR

Issue #36, opened 2021-04-11
assert not implemented correctly on kotlin, go or dart


jayvdb:
```py
def main():
    assert 1 == 1
```
becomes kotlin
```kotlin
fun main() {
assert!(1 == 1)}
```
error:
```kotkin
tests/cases/assert.kt:2:7: error: unexpected tokens (use ';' to separate expressions on the same line)
assert!(1 == 1)
      ^
tests/cases/assert.kt:2:1: error: function invocation 'assert(...)' expected
assert!(1 == 1)
^
tests/cases/assert.kt:2:1: error: none of the following functions can be called with the arguments supplied:
public inline fun assert(value: Boolean): Unit defined in kotlin
public inline fun assert(value: Boolean, lazyMessage: () -> Any): Unit defined in kotlin
assert!(1 == 1)
^
```
go
```go
package todo_naming




func main() {
assert!(1 == 1)}
```
error
```go
tests/cases/assert.go:7:7: syntax error: unexpected ! at end of statement
```
dart:

```dart
 main() {
assert!(1 == 1)}
```

error:
```dart
tests/cases/assert.dart:2:7: Error: Expected '(' before this.
assert!(1 == 1)}
      ^
tests/cases/assert.dart:2:7: Error: Expected ';' after this.
assert!(1 == 1)}
      ^
tests/cases/assert.dart:2:15: Error: Expected ';' after this.
assert!(1 == 1)}
              ^
```

py `assert True` is even worse, as nim and rust fail due to `bool` literals lacking support.

PR #53, opened 2021-04-14
Fixup assert for julia, go, dart and kotlin

adsharma:


jayvdb:
It would be good to get https://github.com/adsharma/py2many/pull/47 in first, so that CI is checking that the asserts actually run.

adsharma:
Yes - these two are relatively minor one off commits. I'll leave them open and wait for #47.

tests/cases/assert.py
```diff
@@ -0,0 +1,2 @@
+def compare_assert(a: int, b: int):
+    assert a == b
```
jayvdb, line 2: all new cases should include a `main()`, so that CI works.

when I add a `main(): compare_assert(1, 2)` (which should fail), dart, julia, kotlin and nim all pass.  Only rust and go fail.  https://github.com/adsharma/py2many/issues/36 should be re-opened.

PR #68, opened 2021-04-16
julia: unmask test failures

adsharma:
Add a compiler, so we can detect out of sync expected outputs

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -17,7 +17,7 @@ COMPILERS = {
     "cpp": ["clang", "-std=c++14"],
     "dart": ["dart", "compile", "exe"],
     "go": ["go", "tool", "compile"],
-    "julia": None,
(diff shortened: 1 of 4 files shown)


jayvdb (approved):

tests/test_cli.py
```diff
@@ -17,7 +17,7 @@
     "cpp": ["clang", "-std=c++14"],
     "dart": ["dart", "compile", "exe"],
     "go": ["go", "tool", "compile"],
-    "julia": None,
+    "julia": ["julia"],
```
jayvdb, line 20: maybe adding "--compiled-modules=yes" like below might reduce the cost of compiling and invoking separately.

The test runner can soon get a rewrite, requiring all new test cases have a "main", and making "INVOKER" the focus of the logic and then "COMPILER" is just an extra step needed for some langs.

PR #71, opened 2021-04-16
Fix AnnAssign for nim and kotlin

jayvdb:


adsharma:
Thanks for fixing. The code looks fine. Perhaps run black/nimpretty on the generated code.

Both of them prefer:

```
a: int = 10

vs

a : int = 10
```

jayvdb:
will do, and will push after rebasing on top of https://github.com/adsharma/py2many/pull/67

PR #84, opened 2021-04-18
ci: Add julia and formatter on macOS and Linux 

jayvdb:
Sits on top of https://github.com/adsharma/py2many/pull/60

### DIFF
--- a/py2many/cli.py
+++ b/py2many/cli.py
@@ -5,6 +5,7 @@ import pathlib
 import subprocess
 
 from dataclasses import dataclass
+from distutils import spawn
 from typing import List, Optional
 
 from .analysis import add_imports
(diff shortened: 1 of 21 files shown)


py2many/cli.py
```diff
@@ -69,7 +69,7 @@ def rust_settings(args):
 
 
 def julia_settings(args):
-    return LanguageSettings(JuliaTranspiler(), ".jl", ["jlfmt"])
+    return LanguageSettings(JuliaTranspiler(), ".jl", ["format.jl"])
```
adsharma, line 72: This is ok as long as you prefix it with ijulia:

```
$ time -p bin/format.jl ~/src/py2many/tests/expected/binit.jl 
real 17.34
user 17.61
sys 0.70
$ time -p ijulia bin/format.jl ~/src/py2many/tests/expected/binit.jl 
real 1.98
user 2.19
sys 0.72
```

jayvdb: Adding a dependency on ijulia is very expensive on CI, and more daunting for newcomers wanting to figure out how to install get pyjl working.

It is easier for you to create your own `format.jl` in your PATH which has your preferred logic.

adsharma: ijulia is not a new package. It's a one line shell script:

```
julia -O0 --compile=min --startup=no $*
```

I tried adding a similar shebang line to format.jl, but it didn't reduce the runtime.

Making CI less expensive in overall runtime is why I'm interested in this.

adsharma: Namespace collision I wasn't aware of. The change is so trivial that it doesn't seem to be worth the complexity of adding a new script.

How about:

```
return LanguageSettings(JuliaTranspiler(), ".jl", ["julia", "-O0", "--compile=min", "--startup=no", "format.jl"])
```

I pasted performance numbers on my machine above. Without these args, formatter is 10x slower on my machine.

PR #87, opened 2021-04-19
pyrs: Use lazy_static for set and dict globals

jayvdb:
rust now support set and dict globals.
dart already supported them.

### DIFF
--- a/pyrs/transpiler.py
+++ b/pyrs/transpiler.py
@@ -6,7 +6,6 @@ from .declaration_extractor import DeclarationExtractor
 
 from py2many.analysis import (
     add_imports,
-    get_element_types,
     get_id,
     is_global,
     is_mutable,
@@ -14,7 +13,7 @@ from py2many.analysis import (
 )
 from py2many.annotation_transformer import add_annotation_flags
 from py2many.context import add_variable_context, add_list_calls
-from py2many.inference import get_inferred_type, is_reference
+from py2many.inference import get_element_types, get_inferred_type, is_reference
 from py2many.mutability_transformer import detect_mutable_vars
 from py2many.scope import add_scope_context
 from py2many.tracer import is_list, defined_before, is_class_or_module
@@ -54,7 +53,9 @@ class RustTranspiler(CLikeTranspiler):
         externs = [f"extern crate {dep};" for dep in deps]
         deps = ",".join(deps)
         externs = "\n".join(externs)
-        uses = "\n".join(f"use {mod};" for mod in usings if mod != "strum")
+        uses = "\n".join(
+            f"use {mod};" for mod in usings if mod not in ("strum", "lazy_static")
+        )
         return f"// cargo-deps: {deps}\n{externs}\n{uses}" if deps else f"{uses}"
 
     def visit_FunctionDef(self, node):
@@ -553,7 +554,7 @@ class RustTranspiler(CLikeTranspiler):
         elif isinstance(node.value, ast.List):
             count = len(node.value.elts)
             elements = ", ".join([self.visit(e) for e in node.value.elts])
-            types = get_element_types(node.value.elts, node.scopes, self._type_map)
+            types = get_element_types(node.value.elts)
             if len(set(types)) == 1 and types[0] is not None:
                 typename = types[0]
             else:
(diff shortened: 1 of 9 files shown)


pyrs/transpiler.py
```diff
@@ -545,6 +546,49 @@ def visit_Assign(self, node):
...
+            else:
+                key_typename = "_"
+            value_types = get_element_types(
+                node.value.values, node.scopes, self._type_map
+            )
+            if len(set(value_types)) == 1 and value_types[0] is not None:
+                value_typename = value_types[0]
+            else:
+                value_typename = "_"
+
```
adsharma, line 579: This logic is already incorporated into type inference. What does `node.annotation` return?

tests/expected/global2.rs
```diff
@@ -0,0 +1,23 @@
...
+extern crate lazy_static;
+use lazy_static::lazy_static;
+use std::collections::HashMap;
+use std::collections::HashSet;
+const code_0: i32 = 0;
+const code_1: i32 = 1;
+const code_a: &str = "a";
+const code_b: &str = "b";
+lazy_static! {
+    static ref l_b: HashSet<&'static str> = [code_a].iter().cloned().collect::<HashSet<_>>();
```
adsharma, line 11: `iter().cloned()` seems to work ok here, but is memory inefficient.

`into_iter()` seems to be more idiomatic code in my searches:

```
// cargo-deps: lazy_static
extern crate lazy_static;
use lazy_static::lazy_static;
use std::collections::HashMap;

lazy_static!{
    static ref MAP: HashMap<&'static str, i32> = vec![
        ("apple", 1),
        ("bear", 2),
        ("cat", 3),
    ].into_iter().collect();
}

fn main() {
    for (k, v) in MAP.iter() {
        println!("{} {}", k, v);
    }
}
```

adsharma: I'm ok with taking the code as-is and figure out how to make it more memory efficient later.

Also, the `iter()` in for loop needs to be brought back if we go this route (conditionally, only if the container is a static ref).

tests/cases/global2.py
```diff
@@ -0,0 +1,18 @@
...
+l_b = {
+    code_a,
+}
+
+l_c = {
+    code_b: code_0,
+}
+
+if __name__ == "__main__":
+    for i in l_b:
```
jayvdb, line 16: This has been broken by ea70634e - the `for` on sets no longer works in rust.
`if "a" in l_b:` still works, and could be used instead here, but that is broken in dart (and all other languages other than rust per https://github.com/adsharma/py2many/issues/93).
If possible, I'd like to keep the test working with a second language, and getting `in` comparison working on dart is one way of doing it.

tests/expected/global2.rs
jayvdb: `iter().cloned()` is an unrelated piece of the transpiler this PR is working on.  I suggest creating an issue about your suggestion here, as it is a good one, but should be looked at separately.

Re-adding `iter()` for specific container types will eventually be need.  The `strum` enums need `::iter`, although there is a issue in that project for adding `.iter()` iirc.  I'll look into this as one option to get this PR ready again.

pyrs/transpiler.py
jayvdb: Now I've got the test case rebuilt, I'll check this out.

jayvdb: `node.annotation` and `node.value.annotation` in the context of visit_Assign didn't have anything for these.
However the implementation of `get_element_types` has now become much simpler.

adsharma:
Would you be ok with merging the first two commits in this PR?

I added some code on top of your third diff that removes `get_element_types` from this code and does type mapping in a more generic way. Also passes the tests you added. Perhaps we can discuss that in a separate PR.

pyrs/transpiler.py
adsharma: The previous version of `get_element_types` could infer that:

```
def foo(a: int):
    c = {}
    d = 1
    if a < 10:
        b = d
        if b > 0:
            c = {b: 3}
 
    return c    
```

c is a `Dict[int, int]`. The variant in this diff can only do that for a limited number of cases (in that it doesn't chase down the definition of `b`).

But I do agree

jayvdb: I do not mind that this `get_element_types` isnt as good as the previous one, especially if you are going to re-write it afterwards.  `get_element_types` is only used to establish whether all members of a global dict/set/list have the same type.  It isnt supposed to be used in any other way.  Your code sample above is not what this function is being used for.

IMO the code only needs to pass the tests that we've established.  Otherwise progress is going to be much slower and more frustrating.

And thus IMO it is counter-productive to merge only the first two commits here - the third commit has the tests for the first two.  The third commit provides the basis on which you can further enhance the type inference, and ideally you add more tests.

adsharma:
Sounds reasonable. I'll go ahead and merge this as-is. The code that builds on top of this is in

https://github.com/adsharma/py2many/pull/98

Issue #19, opened 2021-04-10
Generated binit.rs has lots of mismatched types


jayvdb:
```rust
error[E0308]: mismatched types
 --> src/binit.rs:6:18
  |
6 |     while (low < high) {
  |                  ^^^^ expected `i32`, found `usize`
  |
help: you can convert a `usize` to an `i32` and panic if the converted value doesn't fit
  |
6 |     while (low < high.try_into().unwrap()) {
  |                  ^^^^^^^^^^^^^^^^^^^^^^^^

error[E0308]: mismatched types
 --> src/binit.rs:7:43
  |
7 |         let middle: _ = i32::from(((low + high) / 2));
  |                                           ^^^^ expected `i32`, found `usize`

error[E0277]: cannot add `usize` to `i32`
 --> src/binit.rs:7:41
  |
7 |         let middle: _ = i32::from(((low + high) / 2));
  |                                         ^ no implementation for `i32 + usize`
  |
  = help: the trait `Add<usize>` is not implemented for `i32`

error[E0277]: the type `[i32]` cannot be indexed by `i32`
 --> src/binit.rs:8:19
  |
8 |         if item < data[middle] {
  |                   ^^^^^^^^^^^^ slice indices are of type `usize` or ranges of `usize`
  |
  = help: the trait `SliceIndex<[i32]>` is not implemented for `i32`
  = note: required because of the requirements on the impl of `Index<i32>` for `Vec<i32>`

error[E0308]: mismatched types
 --> src/binit.rs:9:20
  |
9 |             high = middle;
  |                    ^^^^^^ expected `usize`, found `i32`
  |
help: you can convert an `i32` to a `usize` and panic if the converted value doesn't fit
  |
9 |             high = middle.try_into().unwrap();
  |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0277]: the type `[{integer}]` cannot be indexed by `i32`
  --> src/binit.rs:23:9
   |
23 |         bins[bisect_right(limits, d)] += 1;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ slice indices are of type `usize` or ranges of `usize`
   |
   = help: the trait `SliceIndex<[{integer}]>` is not implemented for `i32`
   = note: required because of the requirements on the impl of `Index<i32>` for `Vec<{integer}>`

```

adsharma:
Changes needed to binit.rs to make it pass:

[...]

PR #90, opened 2021-04-19
Rust reference

adsharma:


jayvdb:
ae2e4b7 - using "Partially fixes: https://github.com/adsharma/py2many/issues/19" in your commit message will likely trigger github auto-closing of that issue.

"Fixes" keyword should be for completely fixing the stated issue.

binit.rs still has one reference problem after this:
```rust
error[E0308]: mismatched types
  --> binit.rs:23:36
   |
23 |         bins[bisect_right(&limits, d) as usize] += 1;
   |                                    ^
   |                                    |
   |                                    expected `i32`, found `&i32`
   |                                    help: consider dereferencing the borrow: `*d`
```

re ea70634, your commit message isnt a good justification for reverting my change.  Yes it helps `range()` and a few other cases where `.iter()` isnt implemented, but as shown at https://github.com/adsharma/py2many/issues/91 , `range()` is not something that is something that is supported.  Getting a function like `range()` working only on rust is not a huge win - the cost is that it will break my enhancement https://github.com/adsharma/py2many/pull/87 
```rust
error[E0277]: `l_b` is not an iterator
  --> global2.rs:20:14
   |
20 |     for i in l_b {
   |              ^^^ `l_b` is not an iterator
   |
   = help: the trait `Iterator` is not implemented for `l_b`
   = note: required because of the requirements on the impl of `IntoIterator` for `l_b`
   = note: required by `into_iter`
```

As far as I can see, doing ea70634 is orthogonal to this PR; it doesnt help rust references.  It should be a separate PR which has test cases for extra functionality achieved by removing the `.iter()`.

adsharma:
I have a coverage.py test that passes across all languages and includes range. I consider it basic functionality.
For rust, we have a simple workaround, use references, which the compiler implicitly expands to iter().

Noted about "Partially fixes". Will reopen the issue after merge and use "Related to" in the future.

PR #92, opened 2021-04-20
Improve test coverage

adsharma:
Mainly fixes the python `range` functionality across languages

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -20,7 +20,7 @@ UPDATE_EXPECTED = os.environ.get("UPDATE_EXPECTED", False)
 COMPILERS = {
     "cpp": ["clang++", "-std=c++14", "-I", str(ROOT_DIR), "-stdlib=libc++"],
     "dart": ["dart", "compile", "exe"],
-    "go": ["go", "tool", "compile"],
+    "go": ["go", "build"],
     "julia": ["julia", "--compiled-modules=yes"],
     "kotlin": ["kotlinc"],
     "nim": ["nim", "compile", "--nimcache:."],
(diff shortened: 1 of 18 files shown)


tests/test_cli.py
```diff
@@ -20,7 +20,7 @@
 COMPILERS = {
     "cpp": ["clang++", "-std=c++14", "-I", str(ROOT_DIR), "-stdlib=libc++"],
     "dart": ["dart", "compile", "exe"],
-    "go": ["go", "tool", "compile"],
+    "go": ["go", "build"],
```
adsharma, line 23: Some context about this change

https://github.com/golang/go/issues/32139

jayvdb: I see you are still getting CI error
```
cases/coverage.go:5:2: cannot find package "github.com/hgfischer/go-iter" in any of:
	/opt/hostedtoolcache/go/1.15.11/x64/src/github.com/hgfischer/go-iter (from $GOROOT)
	/home/runner/go/src/github.com/hgfischer/go-iter (from $GOPATH)
```
I havent tried it out, but you might instead install that package in CI before invoking `tox`.  Not ideal, but it may be sufficient to get this working

jayvdb: I see you did that.  Was this change to `COMPILERS` still necessary after that?

I quickly played with using https://github.com/cosmos72/gomacro as the `COMPILER` ; seems like that could be another approach, like we are doing for rust-script and kscript.

adsharma: Yes - still needed. Without it I get:

cases/coverage.go:5:2: can't find import: "github.com/hgfischer/go-iter"

No objections to using something like gomacro. Can re-evaluate if there is a significant difference between the inputs accepted by the compiler vs script (for all languages using an interpreter)

PR #98, opened 2021-04-22
Type inference

adsharma:


adsharma:
One of the main ideas of this pull requests is to make `node.annotation` immutable.

Right now, because it's mutated by various visitors, it's hard to guess if it's a python type, a rust type or some mix of the two. After this change, node.annotation should always be a python type. node.container_type will always be a tuple of python type `(List, int)` for example. All target language specific types are produced via: `_typename_from_annotation`.

Hopefully this will result in an "exactly once" lookup of the type map. Many of the surprising behaviors in the current code base are caused by looking up target specific types (i32 for example) in `self._type_map` and masked via `if var in self._type_map` checks. 

We should do another pass over the code base to fix up any remaining instances that don't go through `_typename_from_annotation` and access the lower level data structures (_type_map and _container_type_map) directly.

py2many/clike.py
```diff
@@ -71,17 +74,57 @@ def _slice_value(self, node: ast.AST):
...
+                return self._map_type(get_id(typename))
+            elif not isinstance(typename, str):
+                # print(f"warning: {typename}")
                 typename = self.visit(typename)
-                typename = self._type_map.get(typename, default_type)
+                return self._map_type(typename)
             else:
-                if typename in self._type_map:
-                    typename = self._type_map[typename]
+                return self._map_type(typename)
```
adsharma, line 130: It's possible that lines 122-127 are bogus and could potentially be removed once we cleanup the rest of the code.

Issue #41, opened 2021-04-12
Dart int_enum does not compile


jayvdb:
```dart
cases/int_enum.dart:6:7: Error: 'RED' is already declared in this scope.
  var RED = auto();
      ^^^
cases/int_enum.dart:2:7: Context: Previous declaration of 'RED'.
  ST0 RED;
      ^^^
cases/int_enum.dart:7:7: Error: 'GREEN' is already declared in this scope.
  var GREEN = auto();
      ^^^^^
cases/int_enum.dart:3:7: Context: Previous declaration of 'GREEN'.
  ST1 GREEN;
      ^^^^^
cases/int_enum.dart:8:7: Error: 'BLUE' is already declared in this scope.
  var BLUE = auto();
      ^^^^
cases/int_enum.dart:4:7: Context: Previous declaration of 'BLUE'.
  ST2 BLUE;
      ^^^^
cases/int_enum.dart:16:7: Error: 'R' is already declared in this scope.
  int R = 1;
      ^
cases/int_enum.dart:12:7: Context: Previous declaration of 'R'.
  ST0 R;
      ^
cases/int_enum.dart:17:7: Error: 'W' is already declared in this scope.
  int W = 2;
      ^
cases/int_enum.dart:13:7: Context: Previous declaration of 'W'.
  ST1 W;
      ^
```

jayvdb:
`str_enum` has the same problems.

adsharma:
@jayvdb - there is a problem with dart pub get vnum. Details in https://github.com/AmirKamali/Flutter_Vnum/issues/7

Do you know if https://github.com/marketplace/actions/flutter-action is usable for CI?

adsharma:
If we manage to land this PR, dart will pass 100% of the tests too.

PR #112, opened 2021-04-24
dart: support enums

adsharma:
State of Enums in Dart (2021)
https://cloud-trends.medium.com/advanced-enums-in-flutter-a8f2e2702ffd

Fixes https://github.com/adsharma/py2many/issues/41

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -59,7 +59,8 @@ jobs:
         run: |
           (cd tests && dart pub global run stagehand:stagehand --override package-simple \
            && dart pub add sprintf \
-           && dart pub add tuple)
+           && dart pub add tuple \
+           && dart pub add vnum)
 
       - name: Install rust-script and cargo-script
         run: cargo install rust-script cargo-script
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,7 +24,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: dart-lang/setup-dart@v1
+      - uses: subosito/flutter-action@v1
+        with:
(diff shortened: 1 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -59,7 +59,8 @@ jobs:
         run: |
           (cd tests && dart pub global run stagehand:stagehand --override package-simple \
            && dart pub add sprintf \
-           && dart pub add tuple)
+           && dart pub add tuple \
+           && dart pub add vnum)
```
jayvdb, line 63: odd. when I run this locally, I dont get an error.  I didn't have flutter installed.  Probably because i previously had done `dart pub add sdk`.  I am guessing that https://github.com/AmirKamali/Flutter_Vnum/issues/7 is going to take a while, as there are quite significant deps in that project on flutter, esp their tests.  IMO, add the extra deps to CI to get this merged (and maybe even check in tests/pubspec.yaml? since the list of deps is growing quickly, then we dont need to install `stagehand`), and note in the readme that wrt Dart, use of IntEnum and StrEnum depends on flutter.

jayvdb (approved):

.github/workflows/main.yml
jayvdb: Possibly only install and run dart tests on one OS, if the deps take too long to install on all of them.

adsharma: I'm not sure if I can get install vnum by adding more dependencies. My guess is that something in dart-lang setup action is forbidding flutter, which is required for vnum.

The only solution I can think of is to use flutter setup action (which  may be more expensive) and use `flutter pub get`

jayvdb: yup, try that.  we can optimise it later

Issue #80, opened 2021-04-17
binit.jl fails to run


jayvdb:
```
ERROR: LoadError: type Array has no field push
Stacktrace:
 [1] getproperty(x::Vector{Int64}, f::Symbol)
   @ Base ./Base.jl:33
 [2] bin_it(limits::Vector{Int64}, data::Vector{Int64})
   @ Main ~/python/py2many/tests/cases/binit.jl:20
 [3] main()
   @ Main ~/python/py2many/tests/cases/binit.jl:82
 [4] top-level scope
   @ ~/python/py2many/tests/cases/binit.jl:85
```

adsharma:
```
-        middle = Int64(((low + high) / 2))
+        middle = Int64(floor((low + high) / 2))
```

Fixes it. Need to convince the transpiler to generate this code based on type inference.

PR #126, opened 2021-04-26
julia: add casts and converts types where necessary

adsharma:
Fixes https://github.com/adsharma/py2many/issues/78
Fixes https://github.com/adsharma/py2many/issues/80


adsharma:
This commit makes Julia the first language to pass all the tests!
cpp and rust are pretty close too.

PR #145, opened 2021-04-29
Go contains

adsharma:
Fixes several tests


adsharma:
Test run is going to fail because 

```
go get github.com/adsharma/py2many/pygo/runtime
```

is commented out. My thought is to land the diff ignoring the failure and immediately send another PR with that line uncommented.

@jayvdb  any other thoughts on how to solve this chicken and egg?

adsharma:
Trying via https://github.com/adsharma/py2many/pull/146

PR #163, opened 2021-05-04
Linting of generated output

jayvdb:


py2many/cli.py
```diff
@@ -215,6 +217,11 @@ def _process_once(settings, filename, outdir):
         if subprocess.call([*settings.formatter, output_path]):
             print(f"Error: Could not reformat: {output_path}")
             return False
+    if settings.ext == ".kt":
+        # ktlint formatter needs to be invoked twice before output is lint free
```
adsharma, line 222: I hope someone writes a better linter for kotlin. It's one of the slower backends already

adsharma (approved):

py2many/cli.py
jayvdb: Meh, we just need to get this fixed.
https://github.com/pinterest/ktlint/issues/1116 seems similar, but not quite the same issue.

jayvdb:
Do you intend to merge this, or https://github.com/adsharma/py2many/pull/160 and then I can rebase.

Issue #167, opened 2021-05-04
Exception in _typename_from_annotation for returns


jayvdb:
Input:
```py
def return_in(code):
    return code in [1, 2, 4]
```

The following occurs for Rust, Nim & Go, at least.  Not cpp.

```py
Traceback (most recent call last):
  File "/usr/local/Caskroom/miniconda/base/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/Caskroom/miniconda/base/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/john.vandenberg/python/py2many/py2many/__main__.py", line 3, in <module>
    main()
  File "/Users/john.vandenberg/python/py2many/py2many/cli.py", line 273, in main
    _process_once(settings, source, outdir)
  File "/Users/john.vandenberg/python/py2many/py2many/cli.py", line 209, in _process_once
    transpile(
  File "/Users/john.vandenberg/python/py2many/py2many/cli.py", line 82, in transpile
    code = transpiler.visit(tree) + "\n"
  File "/Users/john.vandenberg/python/py2many/py2many/clike.py", line 145, in visit
    return super().visit(node)
  File "/usr/local/Caskroom/miniconda/base/lib/python3.8/ast.py", line 363, in visit
    return visitor(node)
  File "/Users/john.vandenberg/python/py2many/pyrs/transpiler.py", line 404, in visit_Module
    buf += [self.visit(b) for b in node.body]
  File "/Users/john.vandenberg/python/py2many/pyrs/transpiler.py", line 404, in <listcomp>
    buf += [self.visit(b) for b in node.body]
  File "/Users/john.vandenberg/python/py2many/py2many/clike.py", line 145, in visit
    return super().visit(node)
  File "/usr/local/Caskroom/miniconda/base/lib/python3.8/ast.py", line 363, in visit
    return visitor(node)
  File "/Users/john.vandenberg/python/py2many/pyrs/transpiler.py", line 77, in visit_FunctionDef
    body = "\n".join([self.visit(n) for n in node.body])
  File "/Users/john.vandenberg/python/py2many/pyrs/transpiler.py", line 77, in <listcomp>
    body = "\n".join([self.visit(n) for n in node.body])
  File "/Users/john.vandenberg/python/py2many/py2many/clike.py", line 145, in visit
    return super().visit(node)
[...]

PR #169, opened 2021-05-04
inference: compare returns bool

adsharma:
Fixes https://github.com/adsharma/py2many/issues/167


jayvdb:
Why not add the sample input to coverage.py or similar, so we dont have regressions.

jayvdb (approved):

PR #173, opened 2021-05-04
Nested container handling

adsharma:


tests/cases/coverage.py
```diff
@@ -10,6 +10,22 @@ def indexing():
     return sum
 
 
+def infer_bool(code: int):
+    return code in [1, 2, 4]
+
+
+# Uncomment when all failures are fixed
```
jayvdb, line 17: Could you move this to a `nested_container.py` or `complex_container.py` or even just `container.py` test case, as we want to be tracking and asserting which languages support this, and which do not.  And there are *lots* of oddballs that can be added to this test file, and getting them all to work in all languages is going to be a very long road to travel.  e.g. adding lambdas, generators, etc to dict values ;-)  And then complex keys.

jayvdb (approved):

PR #184, opened 2021-05-05
Test unsupported features

jayvdb:
Sits on top of https://github.com/adsharma/py2many/pull/186 and https://github.com/adsharma/py2many/pull/185

TODO:
- ~~Remove extra ';' added to nim~~

### DIFF
--- a/py14/transpiler.py
+++ b/py14/transpiler.py
@@ -409,15 +409,6 @@ class CppTranspiler(CLikeTranspiler):
         buf.append("}")
         return "\n".join(buf)
 
-    def visit_Expr(self, node):
-        s = self.visit(node.value)
-        if s.strip() and not s.endswith(";"):
-            s += ";"
-        if s == ";":
-            return ""
-        else:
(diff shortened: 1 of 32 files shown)


jayvdb:
The output of the unsupported skips is often quite ugly.  This can be fixed quite easily by detecting and raising `NotImplementedError`, or by fixing them of course.

adsharma:
Do you know if the regression in CI runtimes is caused by something in this diff stack or something happening on github backends? 25 mins for mac/windows seems high.

jayvdb:
11 mins for invocation of `tox` on Ubuntu.

jayvdb:
Also number of tests increases from 207 -> 389.  Coverage increases 2%, with non-branch missings going from 930 to 820; `py2many/declaration_extractor.py` coverage goes from 66% to 82%

adsharma:
Ah I see. Gives us an incentive to fix all those cases so the tests will be faster :)

PR #185, opened 2021-05-05
CI: Fix golint

jayvdb:
Sits on top of https://github.com/adsharma/py2many/pull/186


tests/test_cli.py
```diff
@@ -44,7 +44,11 @@
     if not item.stem.startswith("test_")
 ]
 
-EXPECTED_COMPILE_FAILURES = []
+EXPECTED_COMPILE_FAILURES = [
```
adsharma, line 47: break this into EXPECTED_COMPILE vs LINT_FAILURES?

Issue #74, opened 2021-04-16
f-string support


jayvdb:
f-strings (`ast.JoinedStr`) are not supported yet.  They should be fairly simple to implement the basics for most languages as string interpolation is available and mostly consistent in the target languages.

Simplest test cases is
```
    var = "Hello world!"
    var2 = f"{var}"
    print(var2)
```
Because it isnt implement, that transpiles to rust
```rust
    let var: &str = "Hello world!";
    let var2: _ = None;
    println!("{}", var2);
```

Using `print(f"{var}")` fails due to logic within the `visit_print` of each transpiler.

I would be happy if either of those two ways of using f-strings worked.  It might be simpler to get them working only in the confines of `visit_print`, where the lifetime of the new string is limited.

jayvdb:
c.f. https://github.com/adsharma/py2many/issues/73 for `str.format` which might be easier to build first.

adsharma:
test case:

```
if __name__ == '__main__':
    a = 10
    print(f"hello world: {a} number")
```

AST looks like this:

```
                           JoinedStr(
                                lineno=3,
                                col_offset=10,
                                end_lineno=3,
                                end_col_offset=36,
                                values=[
                                    Constant(lineno=3, col_offset=10, end_lineno=3, end_col_offset=36, value='hello world: ', kind=None),
                                    FormattedValue(
                                        lineno=3,
                                        col_offset=10,
                                        end_lineno=3,
                                        end_col_offset=36,
                                        value=Name(lineno=3, col_offset=26, end_lineno=3, end_col_offset=27, id='a', ctx=Load()),
                                        conversion=-1,
                                        format_spec=None,
                                    ),
                                    Constant(lineno=3, col_offset=10, end_lineno=3, end_col_offset=36, value=' number', kind=None),
```

I'll prioritize this tomorrow.

jayvdb:
Awesome.  If we can get this fixed, we can transpile `py2many`/`pyrs` itself into Rust - it wont compile, but adding transpiling of py2many as a test case (I have this roughly coded already) will add a lot more coverage preventing exceptions during transpilation.

PR #187, opened 2021-05-05
rewriters: FStringJoinRewriter

adsharma:
This handles python's joined strings as follows:

    print(f"hello world: {a+1} number")

becomes

    println!("{}","".join(vec!["hello world: ", String::from((a + 1)), " number"]););

All we need to do now is to transpile "".join(...) in the target language,
which we need to do anyway because of how popular that construct is in
python code.

Related to: https://github.com/adsharma/py2many/issues/74

### DIFF
--- a/py2many/cli.py
+++ b/py2many/cli.py
@@ -38,7 +38,11 @@ from pygo.transpiler import (
     GoPropagateTypeAnnotation,
 )
 
-from py2many.rewriters import ComplexDestructuringRewriter, PythonMainRewriter
+from py2many.rewriters import (
+    ComplexDestructuringRewriter,
+    FStringJoinRewriter,
+    PythonMainRewriter,
+)
 
 
 def transpile(source, transpiler, rewriters, transformers, post_rewriters):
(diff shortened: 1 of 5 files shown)


jayvdb:
> "".join(...)

This will benefit from some partial fixes in https://github.com/adsharma/py2many/pull/184 re `''.format(..)`.  Or you could take those hints and fix them here, if you're planning on adding tests for this PR.

adsharma:
> > "".join(...)
> 
> This will benefit from some partial fixes in #184 re `''.format(..)`. Or you could take those hints and fix them here, if you're planning on adding tests for this PR.

Can't add tests yet because of

https://github.com/adsharma/py2many/issues/188

adsharma:
> > > "".join(...)
> > 
> > 
> > This will benefit from some partial fixes in #184 re `''.format(..)`. Or you could take those hints and fix them here, if you're planning on adding tests for this PR.
> 
> Can't add tests yet because of
> 
> #188

Resolved that issue. Has a test for rust now. Other languages need "".join() handling help.

PR #189, opened 2021-05-05
tests: make lint configurable

adsharma:
May be useful for testing performance with or without checks.
Also make some of these knobs accessible from cmdline args as
well as env.

Tested via:

PYTHONPATH=. python3 tests/test_cli.py  -k go -v
PYTHONPATH=. python3 tests/test_cli.py --lint -k go -v

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,22 +1,19 @@
+import argparse
 import os.path
 import unittest
 import sys
+
 from distutils import spawn
 from pathlib import Path
 from subprocess import run
 from unittest.mock import Mock
-
 from unittest_expander import foreach, expand
 
 from py2many.cli import main, _get_all_settings
 
-SHOW_ERRORS = os.environ.get("SHOW_ERRORS", False)
-
-TESTS_DIR = Path(__file__).parent
+TESTS_DIR = Path(__file__).parent.absolute()
 ROOT_DIR = TESTS_DIR.parent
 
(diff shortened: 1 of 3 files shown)


tests/test_cli.py
```diff
@@ -64,17 +61,23 @@ class CodeGeneratorTests(unittest.TestCase):
     SETTINGS = _get_all_settings(Mock(indent=4))
     maxDiff = None
 
+    SHOW_ERRORS = os.environ.get("SHOW_ERRORS", False)
+    KEEP_GENERATED = os.environ.get("KEEP_GENERATED", False)
+    UPDATE_EXPECTED = os.environ.get("UPDATE_EXPECTED", False)
+    LINT = os.environ.get("LINT", False)
```
jayvdb, line 67: This should be True by default, especially until the current golint min_confidence 1.0 problems are fixed, otherwise newcomers are going to be surprised when CI fails their tests when they passed locally.

You can however have the CLI disable it by default - newcomers should be running `tox`.

jayvdb: You can also *rename* EXPECTED_COMPILE_FAILURES in this PR to be about lint only, as we now have test_unsupported where we no longer have existing test code which doesnt compile, and new test cases can be partially supported by only checking in some tests/expected/ or by using test_unsupported.

jayvdb (approved):

Issue #7, opened 2021-04-09
Module level docstrings should become comments


jayvdb:

jayvdb:
https://github.com/adsharma/py2many/commit/7f517bee289fcd18a875169c18fc30ea028a6c87 is not the fix.

To explain better, this is not about py2many source - it is when the input Python contains a module level docstring, the equivalent for most languages is a comment, not an unassigned string which is a syntax error in most languages.

jayvdb:
docstrings in functions are also problematic in many of the generated outputs, as they are transpiled to unassigned strings.

adsharma:
Nim has this:

https://github.com/jboy/nim-docstrings

Let's make it a comment by default. If a language has some special support, we can add it in later.

Issue #174, opened 2021-05-04
Go visibility, uppercase first letter


jayvdb:
In order to support directory mode (https://github.com/adsharma/py2many/issues/158), all module level attributes in the Python need to become exposed parts of the Go module, which means rewriting them to use uppercase letter for the first letter.

jayvdb:
For reference in case we need to revisit this at all, go has a tool `gorename` which does this https://pkg.go.dev/golang.org/x/tools/cmd/gorename

PR #190, opened 2021-05-06
rewriter: move docstrings to a AST node comment field

adsharma:
This allows us to consume the rest of the test. The comment
can be generated from the ast node in the target language in
a future diff

Related to: https://github.com/adsharma/py2many/issues/7

### DIFF
--- a/tests/cases/rect.py
+++ b/tests/cases/rect.py
@@ -1,12 +1,16 @@
+"""This file implements a rectangle class """
+
 from dataclasses import dataclass
 
 
 @dataclass
 class Rectangle:
+    """Rectangle as a dataclass"""
     height: int
     length: int
 
     def is_square(self) -> bool:
+        """Go likes this to be camel case"""
         return self.height == self.length
 
 
--- a/tests/test_unsupported.py
+++ b/tests/test_unsupported.py
@@ -72,7 +72,6 @@ TEST_CASES = {
     "percent_formatting": "a = '~ %s ~' % 'a'",  # https://github.com/adsharma/py2many/issues/176
     "tuple_destruct": "foo, (baz, qux) = 4, (5, 6); assert foo != (baz != qux)",  # https://github.com/adsharma/py2many/issues/155
     "float_1": "a = float(1)",  # https://github.com/adsharma/py2many/issues/129
-    "docstrings": "'''docstring'''\ndef main(): pass",  # https://github.com/adsharma/py2many/issues/7
     "print_None": "print(None)",
     "class_vars": "class A:\n  B = 'FOO'\ndef main(): assert A.B == 'FOO'",  # https://github.com/adsharma/py2many/issues/144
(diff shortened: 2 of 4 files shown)


tests/test_unsupported.py
```diff
@@ -72,7 +72,6 @@ class Permissions(IntFlag):
     "percent_formatting": "a = '~ %s ~' % 'a'",  # https://github.com/adsharma/py2many/issues/176
     "tuple_destruct": "foo, (baz, qux) = 4, (5, 6); assert foo != (baz != qux)",  # https://github.com/adsharma/py2many/issues/155
     "float_1": "a = float(1)",  # https://github.com/adsharma/py2many/issues/129
-    "docstrings": "'''docstring'''\ndef main(): pass",  # https://github.com/adsharma/py2many/issues/7
```
jayvdb, line 75: When removing anything from here, please add the same syntax into one of tests/cases/ to verify it is working correctly and prevent regressions.

If you also add docstrings to cases rect.py and *_enum.py , you should be able to remove some/all of those cases from the ignore list, because that is what golint is whining about.

jayvdb (changes requested):

tests/cases/grumpy_ops_test.py
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env python3
+
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
```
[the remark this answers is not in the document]
adsharma: Fair enough. I'll remove the test.

tests/cases/rect.py
```diff
@@ -1,12 +1,16 @@
+"""This file implements a rectangle class """
+
 from dataclasses import dataclass
 
 
 @dataclass
 class Rectangle:
+    "Rectangle as a dataclass"
```
jayvdb, line 8: can we use triple docstrings, per [PEP8](https://www.python.org/dev/peps/pep-0008/#id33)/[PEP257](https://www.python.org/dev/peps/pep-0257/)

tests/cases/rect.py
```diff
@@ -1,12 +1,16 @@
...
     height: int
     length: int
 
     def is_square(self) -> bool:
+        "Go likes this to be camel case"
```
jayvdb, line 13: Could mention https://github.com/adsharma/py2many/issues/174 here.

Are any of the *_enum cases able to be accepted by golint by adding docstrings, or can one be adjusted to pass golint?  The enum values could have the case of their names changed - shouldnt effect anything  else.

jayvdb (changes requested):

jayvdb (approved):

PR #200, opened 2021-05-07
dart: implement exceptions

adsharma:
Other languages which support exceptions can replicate this.


tests/expected/exceptions.dart
```diff
@@ -0,0 +1,26 @@
+// @dart=2.9
```
jayvdb, line 1: tests/case/exceptions.py not included in PR

adsharma: Added now. I need to clean out my checked out git repo so there is less noise.

Issue #21, opened 2021-04-10
Run custom/multiple formatters


jayvdb:
Running additional formatters may catch more syntax problems without invoking the compiler.

Also users may wish to use a different formatter / error detector.

e.g. https://github.com/uncrustify/uncrustify might be an extra formatter for many C-like languages.

jayvdb:
https://sparkbox.com/foundry/go_vet_gofmt_golint_to_code_check_in_Go shows why multiple formatters/linters should be used (in addition to compiling) to ensure the generated code is accurate.  Not very important yet as the go code is rarely compilable atm, but we'll get there very soon.

jayvdb:
c.f. https://github.com/adsharma/py2many/issues/191 wrt go formatters/linters.

PR #214, opened 2021-05-12
cli: Support CXX and CXXFLAGS

jayvdb:
Adds plumbing for using with g++ only, testing with astyle as
an alternative formatter.

Also allow CLANG_FORMAT_STYLE to determine the clang-format
style to be used.

### DIFF
--- a/py2many/cli.py
+++ b/py2many/cli.py
@@ -2,10 +2,10 @@ import argparse
 import ast
 import os
 import pathlib
-import subprocess
 
 from dataclasses import dataclass, field
 from distutils import spawn
+from subprocess import run
 from typing import Callable, List, Optional
 
 from .analysis import add_imports
@@ -119,6 +119,13 @@ def transpile(source, transpiler, rewriters, transformers, post_rewriters):
     return "\n".join(out)
 
 
+def _create_cmd(parts, filename):
+    cmd = [arg.format(filename=filename) for arg in parts]
+    if cmd != parts:
+        return cmd
+    return [*parts, str(filename)]
+
+
 @dataclass
 class LanguageSettings:
     transpiler: CLikeTranspiler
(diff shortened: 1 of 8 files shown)


py2many/cli.py
```diff
@@ -252,19 +275,27 @@ def _process_once(settings, filename, outdir):
...
-            print(f"Error: Could not reformat: {output_path}")
-            return False
-    if settings.ext == ".kt":
-        # ktlint formatter needs to be invoked twice before output is lint free
-        if subprocess.call([*settings.formatter, output_path]):
-            print(f"Error: Could not reformat: {output_path}")
+        cmd = _create_cmd(settings.formatter, output_path)
+        proc = run(cmd, env=env)
+        # format.jl exit code is unreliable
+        if settings.ext == ".jl":
```
jayvdb, line 283: this should be moved inside the `if proc.returncode:` below after the `print`

tests/test_cli.py
```diff
@@ -192,6 +216,69 @@ def test_cli(self, case, lang):
...
+        if not spawn.find_executable(env["CXX"]):
+            raise unittest.SkipTest(f"{env['CXX']} not available")
+
+        settings = _get_all_settings(Mock(indent=4), env=env)[lang]
+        assert settings.linter[0].startswith("g++")
+
+        if not spawn.find_executable("astyle"):
+            raise unittest.SkipTest("astyle not available")
+
+        settings.formatter = ["astyle"]
```
adsharma, line 241: Looks like astyle is an alternative formatter and has nothing to do with whether CXX=gcc or clang.

adsharma (approved):
Good to know that generated code works with g++ as well!

Let me know if you want to rebase it yourself.

tests/test_cli.py
jayvdb: I'd like to remove the dependency on llvm, and clang-format is part of that goal. c.f. https://github.com/adsharma/py2many/issues/21  I guess adding a `--formatter=...` and `--linter=...` cli args is how we would expose that.  For now using astyle here ensures this group of tests doesnt use llvm.

PR #216, opened 2021-05-12
cli: make pypi package work again

adsharma:
pypi package errors out because of env=None.

### DIFF
--- a/py2many/cli.py
+++ b/py2many/cli.py
@@ -303,7 +303,7 @@ def _process_once(settings, filename, outdir, env=None):
     return True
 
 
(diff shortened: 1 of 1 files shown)


adsharma:
@jayvdb I uploaded 0.2 to pypi already. Should've tested a bit more.

Please review and let me know if you can think of any method other than uploading 0.2.1 after this.

Repro: 

```
py2many.py --rust=1 test.py
```

jayvdb (approved):

jayvdb:
Sorry for that. Unfortunately, uploading a new release will be needed.

PR #223, opened 2021-05-14
Bitops

adsharma:


tests/cases/fib.py
```diff
@@ -1,11 +1,14 @@
...
     return fib(i - 1) + fib(i - 2)
 
 
 if __name__ == "__main__":
-	# Using a variable to workaround
-	# https://github.com/adsharma/py2many/issues/64
-	rv = fib(5)
-	print(rv)
+    # Using a variable to workaround
+    # https://github.com/adsharma/py2many/issues/64
```
jayvdb, line 12: this is solved; perhaps just replace all this with `print(fib(5))`

tests/cases/bitops.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python3
```
jayvdb, line 1: no `tests/expected/bitops.*` added in this PR so far.

Issue #179, opened 2021-05-05
transmutation of `getopt`/`argparse`


jayvdb:
Single file Python scripts is likely to be a use-case which will be mostly functional soon/first, but they usually use `argparse`.   Luckily there is usually a similar library in most languages that we can convert the Python into.  `getopt` is likely to be easier to get working in all languages, and it is worth doing for that reason alone, however it is less used in real Python code this days.  But it Python authors might be willing to downgrade their `argparse` code into `getopt` if it means they can transpile to a larger number of languages.

adsharma:
Yes, argparse -> structopt is a useful initial attempt that could help people writing quick python CLIs and getting a native binary out of it easily.

Having said that I'm concerned about such translation adding a lot of cognitive overhead to pyrs/transpiler.py. We need to figure out a clean interface - some sort of a plugin mechanism for rewriters + dispatch_map.

adsharma:
This is not of much help for existing code, but for new code, I find this more appealing:

https://github.com/mivade/argparse_dataclass/blob/master/tests/example.py

```
from argparse_dataclass import dataclass

@dataclass
class Opt:
    x: int = 42
    y: bool = False
```

PR #236, opened 2021-05-20
rust: implement plugin based code generation

adsharma:
This allows us to keep complex use cases such as argument parsing
separate from the core transpiler.

Related to: https://github.com/adsharma/py2many/issues/179

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,7 +43,7 @@ jobs:
       - name: Install tox
         run: pip3 install tox
 
-      - name: Install ${{ runner.os }} dependencies.
+      - name: Install Julia dependencies for ${{ runner.os }}.
         shell: bash
         run: |
           julia -e 'using Pkg; Pkg.add("JuliaFormatter")'
--- a/py2many/clike.py
+++ b/py2many/clike.py
@@ -3,6 +3,7 @@ import importlib
 import logging
 import sys
 
+from pathlib import Path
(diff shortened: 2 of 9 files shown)


.github/workflows/main.yml
```diff
@@ -40,10 +43,17 @@ jobs:
...
-      - name: Install tox
-        run: pip3 install tox
+      - name: Install Python packages
+        shell: bash
+        # Use argparse_dataclass from pypi once a new package has been uploaded
+        run: |
+          (pip3 install tox \
+           && git clone https://github.com/mivade/argparse_dataclass/ \
+           && (cd argparse_dataclass; python setup.py install) \
+           && rm -rf argparse_dataclass \
```
adsharma, line 53: @jayvdb - this isn't working as intended. I suppose tox still can't find `argparse_dataclass`.

Ideally upstream uploads a new package to pypi and we pick it up via `pip3 install`. Do you have any suggestions on the right way to do this?

jayvdb: in tox.ini, add `git+https://github.com/mivade/argparse_dataclass/` to the deps

adsharma:
```
fatal error: 'vector' file not found
```

This looks like a new error that's unrelated to the diff in question.

jayvdb:
I have a fix for `vector` and a bunch of related cpp include improvements.  I'll have it pushed up for review in approx 2 hrs.

PR #241, opened 2021-05-21
pycpp: Use Google style and cpplint

jayvdb:
std includes should be sorted and de-duped.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         os:
-          - ubuntu-20.04
+          - ubuntu-18.04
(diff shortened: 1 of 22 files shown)


jayvdb:
If you're not sure about any of the latter commits, feel free to merge the earlier commits.

.github/workflows/main.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         os:
-          - ubuntu-20.04
+          - ubuntu-18.04
```
adsharma, line 20: Hopefully this is a temporary workaround and we can get back on to 20.04 once this is fixed.

PR #243, opened 2021-05-22
rust: improve plugin system for file IO

adsharma:
Introduces a companion library: pylib. This could be a submodule.
Gets us closer to with_open.py example in CI

Introduces a few new concepts:

MODULE_DISPATCH_TABLE - this is where you map module names (tempfile.xxx)
DECORATOR_DISPATCH_TABLE - tells the transpiler what the decorator is doing
FUNCTION_DISPATCH_TABLE - looked up on attributes as well as func dispatch

### DIFF
--- a/pyrs/plugins.py
+++ b/pyrs/plugins.py
@@ -1,8 +1,11 @@
+import io
+import ast
 import textwrap
 
 try:
     from argparse_dataclass import dataclass as ap_dataclass
     from argparse_dataclass import ArgumentParser
+    from tempfile import NamedTemporaryFile
 except:
     ArgumentParser = "ArgumentParser"
     ap_dataclass = "ap_dataclass"
@@ -45,10 +48,15 @@ class RustTranspilerPlugins:
             self._usings.add("std::fs::OpenOptions")
             mode = vargs[1]
             opts = "OpenOptions::new()"
+            is_binary = "b" in mode
             for c in mode:
                 if c == "w":
+                    if not is_binary:
+                        self._usings.add("pylib::FileWriteString")
                     opts += ".write(true)"
                 if c == "r":
+                    if not is_binary:
+                        self._usings.add("pylib::FileReadString")
                     opts += ".read(true)"
                 if c == "a":
                     opts += ".append(true)"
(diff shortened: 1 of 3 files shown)


pyrs/plugins.py
```diff
@@ -45,10 +48,15 @@ def visit_open(self, node, vargs):
             self._usings.add("std::fs::OpenOptions")
             mode = vargs[1]
             opts = "OpenOptions::new()"
+            is_binary = "b" in mode
             for c in mode:
                 if c == "w":
+                    if not is_binary:
+                        self._usings.add("pylib::FileWriteString")
```
jayvdb, line 55: while crate name `pylib` is available, I don't think that accurately describes the purpose.  `py2many-helpers` is as good a name as I can think of, and is a name which could be used for similar libraries in other languages.

jayvdb: Also a name like this implies that nobody should rely on its API, which I guess will fluctuate as the needs of py2many evolve, and I expect some parts of it might become separate crates when components become mature.

adsharma: Someone could use such things without using py2many right? nim's pylib existed before py2many.

jayvdb: Re "This could be a submodule." I would rather we didnt go down that path.  IMO (2 cents worth) it makes sense to have the helpers as part of this repo, or using cargo's git repo support for libraries not on crates.io, and then as it stablises it can be a published crate on crates.io.  Obviously your call, and not a big deal either way.

jayvdb: Re nim pylib, its stated goal seems broader than you are likely to want to entertain/maintain. c.f. https://github.com/Yardanico/nimpylib/issues/11
And say for example that we put some strenum voodoo into rust pylib, because it is missing in strum, what would happen if strum adds that feature -- I would hope that the rust pylib removes that extra functionality as it is no longer needed.  otherwise the rust pylib becomes a huge library full of things people depend on that py2many no longer needs.  Just my 2c.

adsharma: Yes - submodules are a pain to maintain. Perhaps fetching from crates.io is a better option.

> I would hope that the rust pylib removes that extra functionality as it is no longer needed.

What you're saying is that the bar should be higher for something to be added to pylib. Rust doesn't have an ABI guarantee yet. So what remains is marking things stable/unstable, so users have a hint about API stability.

Looking at nim pylib's list - things like hasattr/getattr/isinstance and anything that depends on runtime types is out for us. Some like pow are covered by  AST transformers. Hopefully it's a much smaller list of frequently used python idioms.

Issue #248, opened 2021-05-22
dict keys()


jayvdb:
Implicit `.keys()` fails for Dart, Julia & Rust.
```py
def implicit_keys():
    CODES = {"KEY": 1}
    return "KEY" in CODES
```

Only Julia can handle explicit `.keys()`
```py
def explicit_keys():
    CODES = {"KEY": 1}
    return "KEY" in CODES.keys()
```

jayvdb:
fyi, I am making good headway on this one.  Will have a PR up in a few hours.

jayvdb:
https://github.com/adsharma/py2many/pull/249 added a test case and a model for how to support .keys() and .values() for Julia and Rust.  Replicating that for other languages should be a good first issue for any newcomers to the project.

PR #249, opened 2021-05-22
dict .keys() and .values()

jayvdb:
Related to https://github.com/adsharma/py2many/issues/248

### DIFF
--- a/pyjl/transpiler.py
+++ b/pyjl/transpiler.py
@@ -231,6 +231,25 @@ class JuliaTranspiler(CLikeTranspiler):
         bytes_str = "{0}".format(node.s)
         return bytes_str.replace("'", '"')  # replace single quote with double quote
 
+    def visit_Compare(self, node):
+        left = self.visit(node.left)
+        right = self.visit(node.comparators[0])
+
+        if hasattr(node.comparators[0], "annotation"):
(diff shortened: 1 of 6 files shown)


pyjl/transpiler.py
```diff
@@ -231,6 +231,21 @@ def visit_Bytes(self, node):
         bytes_str = "{0}".format(node.s)
         return bytes_str.replace("'", '"')  # replace single quote with double quote
 
+    def visit_Compare(self, node):
+        left = self.visit(node.left)
+        right = self.visit(node.comparators[0])
+
+        comparator_typename = self._typename_from_annotation(node.comparators[0])
+        if comparator_typename and comparator_typename.startswith("Dict"):
```
jayvdb, line 239: im not keen on these string comparisions.
I see `hasattr(node, "container_type")` elsewhere .. maybe that will work here.

adsharma: yeah - `container_type` is much more robust. Let's go with that.

adsharma: `*generic_container_type` if you're comparing to `Dict`

jayvdb: Done. ping @adsharma

PR #251, opened 2021-05-22
pyrs: Always use &'static str

jayvdb:


pyrs/inference.py
```diff
@@ -13,7 +13,7 @@
     int: "i32",
     float: "f64",
     bytes: "&[u8]",
-    str: "&str",
+    str: "&'static str",
```
adsharma, line 16: This works when all strings are compile time constants, which is the case for our tests, but not real world programs.

```
#!/usr/bin/env python3

MYTABLE = {
  "k1", "v1",
  "k2", "v2",
}

def lookup(k: str) -> str:
    return MYTABLE[k]
   
if __name__ == '__main__':
    x = sys.stdin.read()
    print(lookup(x))
```

I've been thinking about adding `node.lifetime` parallel to `node.annotation`. This way, we don't pollute annotations with lifetimes for languages that don't care about them.

The ones which do, can make use of the extra info to derive lifetimes.

These extra attributes can over time pile up. So we should document them somewhere. Ones of the top of my head:

* node.scopes
* node.container_types
* node.generic_container_types
* node.declarations
* node.class_assignments

These are similar to rust extension traits.

jayvdb: Granted this PR is a quick fix to get compile time constant strings working in Rust, but it is not a hack.  In reality strings in Python have all the properties of static lifetime: two strings loaded from different sources have the same identity if they have the same content.

`sys.stdin` is not supported yet, but you do some some file io working.  `tests/cases/with_open.py` needs some functions added, like your example above, and perform string comparisons, otherwise it is also very limited and not effected by this change.  Then dynamically loaded strings need to match the semantics of Python strings.

adsharma: Two thoughts here:

One is to implement support for user defined types. Example a plugin.py which allows:

```
RUST_TYPE_MAP = {
  "str": "&mystr",
  ...
}
```

at startup we load the plugin.py file via a command line arg and update the builtin `RUST_TYPE_MAP`.

The other option is to actually implement support for lifetimes in the AST. Let me put up a simple version for discussion later today.

jayvdb:
943bbb3ea means this workaround doesnt work any more, and would force me to create what looks like non-trivial lifetime inference, or create an even more ugly workaround.  My project is going to have to fork off based on this PR until I can figure this out.

adsharma:
> [943bbb3](https://github.com/adsharma/py2many/commit/943bbb3ead1b974eb8a464e995d05f75bcc1cfd2) means this workaround doesnt work any more, and would force me to create what looks like non-trivial lifetime inference, or create an even more ugly workaround. My project is going to have to fork off based on this PR until I can figure this out.

It's a bit unfortunate - but having a static lifetime for all strings can't possibly be correct.

PR #254, opened 2021-05-23
py2many: implement support for checking fixed width types

adsharma:
This is so that we can catch problems at source and report
the lineno instead of relying on the compiler for the target

    a1: i8 = 300   # type error, bit lengths don't match
    a: i8 = 10     # ok
    b: i16 = 300   # ok
    c: i16 = a + b # type error due to potential overflow

### DIFF
--- a/py2many/inference.py
+++ b/py2many/inference.py
@@ -49,6 +49,32 @@ def is_reference(arg):
     return annotation_has_ref
 
 
+def bit_length(val: ast.AST) -> int:
+    if isinstance(val, ast.Constant) and isinstance(val.value, int):
+        return int.bit_length(val.value)
+    return 0
+
+
+def is_compatible(
+    cls1, cls2, target: Optional[ast.AST] = None, source: Optional[ast.AST] = None
+):
+    """This needs to return true if narrowing one of the classes leads to the other class"""
+    # For now, handle fixed width only. In the future look into List[int] vs List
+    fixed_width = InferTypesTransformer.FIXED_WIDTH_INTS
+    fixed_width_bit_length = InferTypesTransformer.FIXED_WIDTH_BIT_LENGTH
+    if cls1 in fixed_width and (cls2 in fixed_width or cls2 is int):
+        target_bit_length = fixed_width_bit_length[cls1]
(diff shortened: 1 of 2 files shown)


py2many/inference.py
```diff
@@ -204,8 +243,21 @@ def visit_AnnAssign(self, node: ast.AnnAssign) -> ast.AST:
...
             self.has_fixed_width_ints = True
+        annotation = get_inferred_type(node.value)
+        value_typename = self._clike._generic_typename_from_type_node(annotation)
+        target_class = class_for_typename(target_typename, None)
+        value_class = class_for_typename(value_typename, None)
+        if (
+            not is_compatible(target_class, value_class, target, node.value)
+            and target_class != None
+        ):
+            raise Exception(
```
jayvdb, line 258: It would be nice if this raised `NotImplementedError` and had a snippet in `test_unsupported` which triggers this.

adsharma: Agreed. Also `node.lineno` is good information to add to all the `NotImplementedError` in the code base so the author of python can quickly track down where the error is coming from.

adsharma: Actually `TypeError` makes more sense here, since there is nothing we can do to implement this.

tests/test_unsupported.py
```diff
@@ -270,6 +277,39 @@ def test_snippet(self, case, lang):
...
+        ext = settings.ext
+        source_data = case
+        expected_error = TEST_ERROR_CASES[case]
+        is_script = has_main(source_data)
+        if ext in [".dart", ".kt", ".rs"] and not is_script:
+            source_data = f"def main():\n  {source_data}"
+        print(f">{source_data}<")
+        tree = ast.parse(source_data)
+        astpretty.pprint(tree)
+        proc = run([sys.executable, "-c", source_data], capture_output=True)
```
jayvdb, line 298: this is not exercising the code to prove it is usable Python code.
and these cases are not usable.
```py
python -c "a: i8 = 10; b: i16 = 300; c: i16 = a + b"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
NameError: name 'i8' is not defined
```

PR #256, opened 2021-05-23
Add sys.argv support within __main__

jayvdb:
Related to https://github.com/adsharma/py2many/issues/179

### DIFF
--- a/py2many/rewriters.py
+++ b/py2many/rewriters.py
@@ -138,7 +138,7 @@ class PythonMainRewriter(ast.NodeTransformer):
             if hasattr(node, "scopes") and len(node.scopes) > 1:
                 rename(node.scopes[-2], "main", "main_func")
             # ast.parse produces a Module object that needs to be destructured
-            ret = ast.parse("def main(): True").body[0]
+            ret = ast.parse("def main(argc: int, argv: List[str]): True").body[0]
             ret.lineno = node.lineno
             ret.body = node.body
             # So backends know to insert argc, argv etc
--- a/pyrs/transpiler.py
+++ b/pyrs/transpiler.py
@@ -158,11 +158,16 @@ class RustTranspiler(CLikeTranspiler):
             del args[0]
             args_list.append("&self")
 
+        is_python_main = getattr(node, "python_main", False)
+
         typedecls = []
         index = 0
         for i in range(len(args)):
             typename = typenames[i]
             arg = args[i]
+            if is_python_main and arg in ["argc", "argv"]:
+                continue
+
             if typename == "T":
                 typename = "T{0}".format(index)
                 typedecls.append(typename)
@@ -255,6 +260,11 @@ class RustTranspiler(CLikeTranspiler):
         attr = node.attr
 
         value_id = self.visit(node.value)
+
+        if value_id == "sys":
+            if attr == "argv":
+                self._usings.add("std::env")
+                return "env::args().map(|s| &*Box::leak(s.into_boxed_str())).collect()"
 
         if is_list(node.value):
             if node.attr == "append":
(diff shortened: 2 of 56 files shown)


pyrs/transpiler.py
```diff
@@ -256,6 +261,11 @@ def visit_Attribute(self, node):
 
         value_id = self.visit(node.value)
 
+        if value_id == "sys":
+            if attr == "argv":
+                self._usings.add("std::env")
+                return "env::args().map(|s| &*Box::leak(s.into_boxed_str())).collect()"
```
adsharma, line 267: What's the Box::leak doing? Do you have a reference for this method?

py2many/rewriters.py
```diff
@@ -138,7 +138,7 @@ def visit_If(self, node):
             if hasattr(node, "scopes") and len(node.scopes) > 1:
                 rename(node.scopes[-2], "main", "main_func")
             # ast.parse produces a Module object that needs to be destructured
-            ret = ast.parse("def main(): True").body[0]
+            ret = ast.parse("def main(argc: int, argv: List[str]): True").body[0]
```
adsharma, line 141: Is pycpp the only one which uses argc? If so, we can probably simplify a bit by removing argc here and handling it inside pycpp,

pyrs/transpiler.py
```diff
@@ -256,6 +261,11 @@ def visit_Attribute(self, node):
 
         value_id = self.visit(node.value)
 
+        if value_id == "sys":
```
adsharma, line 264: Move this to pyrs/plugins.py?

Make a new entry into FUNC_DISPATCH_TABLE:

{ sys.argv: visit_sys_argv }

adsharma (commented):
Much needed functionality!

py2many/rewriters.py
jayvdb: Not it isnt.

pyrs/transpiler.py
jayvdb: https://doc.rust-lang.org/std/boxed/struct.Box.html#method.leak

This is needed to force env::args into a Vec of `str`, which is forced on us at the moment.

pyrs/transpiler.py
jayvdb: It is a common pattern in the visit_Attribute of every transpiler.py , while plugins is pyrs only.

py2many/rewriters.py
adsharma:
```
$ grep argc tests/expected/sys_argv.* -l
tests/expected/sys_argv.cpp
```

which other backend is using argc?

sys_argv* tests pass for me even after removing `argc` in the line above. What breaks if we remove it from all the py*/transpiler.py also in this diff?

pyrs/transpiler.py
adsharma: Fair enough. We can do this in future diffs which introduce plugins.py to all languages. I was trying to keep py*/transpiler.py as focused as possible and move the API translation to plugins and other files.

pyrs/transpiler.py
adsharma:
```
    let __temp1: Vec<String> = env::args().collect();
    let a: Vec<&str> = __temp1.iter().map(|s| s as &str).collect();
```

reads easier. This is fine for now, since generating the above requires additional code in the transpiler.

py2many/rewriters.py
jayvdb: Ah, backend, yes, none are using it currently.

However many languages have the ability to implement the C `main` directly.
e.g. nim has `proc cmain(c: cint, a: cstringArray): cint {.importc.}`.

My rust voodoo avoided me needing to do that, but we probably will want to switch to it for Rust to avoid the `Box::leak`.

Creating a new signature for `main` doesnt make sense.  There is already one global definition of the `main` entry point: the C one, which exists in every program on nearly every OS out there.

adsharma: Every language has it's own signature for main. Most of them (6 out of 7) different from C and they have their own way to map/interop with C main if necessary. Rust has concepts like: `#[start]` that may map to what you're thinking.

https://github.com/rust-lang/rust/pull/5809

When we transpile, we map python's main signature to the target language's main signature, which is what your code is doing.

This isn't a show stopper for me. Since you have other stuff on top of this, let's get this in and work on simplifying the code in later diffs.

PR #259, opened 2021-05-23
tests: Transpile py2many into rust

jayvdb:
This exercises the py2many and pyrs modules with a lot more AST, catching many more types of syntax, hopefully reducing the regressions caused by new features.

I mentioned it previously at https://github.com/adsharma/py2many/issues/74#issuecomment-832336957

### DIFF
--- a/py2many/clike.py
+++ b/py2many/clike.py
@@ -444,7 +444,8 @@ class CLikeTranspiler(ast.NodeVisitor):
     def visit_AnnAssign(self, node):
         target = self.visit(node.target)
         if (
-            isinstance(node.target.annotation, ast.Subscript)
+            hasattr(node.target, "annotation")
+            and isinstance(node.target.annotation, ast.Subscript)
             and get_id(node.target.annotation.value) == "Callable"
         ):
             type_str = self._default_type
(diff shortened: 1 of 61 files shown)


jayvdb:
ping @adsharma , I have a long series here

jayvdb:
Rebased and thankfully green. ping @adsharma

jayvdb:
This increased coverage from 85% to 89%

py2many/clike.py
```diff
@@ -69,12 +69,15 @@
 def class_for_typename(typename, default_type, locals=None) -> Union[str, object]:
     if typename is None:
         return None
+    if typename == "super" or typename.startswith("super()"):
```
adsharma, line 72: Why not add `RuntimeError` to the list below?

adsharma (approved):

py2many/clike.py
jayvdb: Because catching `RuntimeError` is about as bad as catching `BaseException` IMO.

To be honest, IMO, the exception list here is growing too long, and this `eval` is called way too often on obviously rubbish values.  If we're a bit more selective, the exception list could be shorter.  But this is not a high priority.

PR #262, opened 2021-05-23
main: simplify the method signature

adsharma:


jayvdb:
Please no.  My design is not broken, as I explained in https://github.com/adsharma/py2many/pull/256#discussion_r637569301 .  The C entry point is the common interface.  We will need to use `proc cmain` and `#[start]` and others when we implement more features, mark my words.  We will be implementing languages other than C++ which do require `argc` in their main signature.

jayvdb:
A better way to achieve this is for each language to declare whether they need `argc` or `argv`, so the generic rewriter can use that info to generate the appropriate generic signature.  I could put up a PR to demo that if you're not 100% committed to your PR.

adsharma:
> A better way to achieve this is for each language to declare whether they need `argc` or `argv`, so the generic rewriter can use that info to generate the appropriate generic signature. I could put up a PR to demo that if you're not 100% committed to your PR.

Yeah - this seems preferable. The implementation could be one PythonMainRewriter per language.

I'll put this on hold until I understand your idea.

jayvdb:
This can be closed now ?

jayvdb:
ping @adsharma

adsharma:
Covered by  other PRs

Issue #258, opened 2021-05-23
Target lang detection in Python source code


jayvdb:
It would be *very* helpful when building test cases if we can use logic like the following to exclude some target languages:

```py
if not sys.argv[0].endswith(".kt"):
   do_something_which_breaks_only_on_kotlin()
```
That depends on https://github.com/adsharma/py2many/pull/256 and https://github.com/adsharma/py2many/issues/257

Another approach is the set a special global like `sys.implementation` to various values like `py2many-kotlin`,

adsharma:
It's a sharp knife - just like `ifdef` in C.

What's the argument against breaking that line/stanza into it's own test which doesn't have an expected result for $lang (i.e. not checkin test_expected.kt for this case). It's extra work to make smaller test cases, true. But it has some benefits too.

jayvdb:
The arguments against it are

1. we dont do that.  we could, but we dont, because it is ugly
2. it is more natural and better organised to keep test cases about functional areas

I definitely don't want to invent a language on top of python, so we're not going near `#ifdef` territory.

What should `print(sys.implementation)` do in generated code?

adsharma:
name=py2many version=pypi-version?

adsharma:
Ah I see what you're asking: we could include target=rust as well.

jayvdb:
Sort-of.  I am not interested in creating a new field like `target`.

`python -c 'import sys;print(sys.implementation.name)'` = `cpython`

py2many generated code is not running on `cpython`.  So it should be something like `py2many`.  But `pyrs` is not the same as `pyjl`.  So I propose .name = `py2many-pyrs`, etc

PR #263, opened 2021-05-23
Substr search

adsharma:


py2many/rewriters.py
```diff
@@ -262,3 +263,28 @@ def visit_Call(self, node):
...
+        if self._language in {"dart", "kotlin"}:
+            return node
+
+        if isinstance(node.ops[0], ast.In):
+            left = node.left
+            right = node.comparators[0]
+            left_type = get_id(get_inferred_type(left))
+            right_type = get_id(get_inferred_type(right))
+            if left_type == "str" and right_type == "str":
+                ret = ast.parse("a.find(b) != -1").body[0].value
```
jayvdb, line 283: Before settling on one canonical way to achieve str-in-str, I'd like to ensure we can easily implement that in each language.  To be useful for https://github.com/adsharma/py2many/issues/258, it must work in all languages.

The reason I dont like this approach is that the `-1` on the rhs is a magic value defined by the function on the lhs.  Each language is going to need to detect `.find` and then replace `-1` with whatever their function returns.

IMO it is better for each language to handle `'a' in 'ba'` directly and implement it using the best "contains" routine that can produce a boolean.  We already have one `if self._language` in `py2many/`, and IMO we should not add more.

---

Re `str.find`, Nim returns -1.  I am surprised that the str2str test case should have reported an error on this PR, as this change here should be sufficient to cause nim to need to be added to the `EXPECTED_SUCCESSES`.

Go's equivalent is strings.Index, and uses -1.

For C++, str.find returns `string::npos`.  If that is somehow handled by py2many to be `-1`, it should work. 

Julia has https://docs.julialang.org/en/v1/base/strings/index.html#Base.findfirst-Tuple{AbstractString,%20AbstractString} and it looks like the value to expect is `nothing`.  https://docs.julialang.org/en/v1/base/strings/#Base.occursin would be easier to implement as it returns bool.

On Rust, str.find uses `Result`, and that doesn't work in py2many yet.  `str.contains` avoids that headache as it returns bool.

py2many/rewriters.py
```diff
@@ -262,3 +263,43 @@ def visit_Call(self, node):
...
+    def visit_Compare(self, node):
+        if self._language in {"dart", "kotlin", "nim"}:
+            return node
+
+        if isinstance(node.ops[0], ast.In):
+            left = node.left
+            right = node.comparators[0]
+            left_type = get_id(get_inferred_type(left))
+            right_type = get_id(get_inferred_type(right))
+            if left_type == "str" and right_type == "str":
```
jayvdb, line 282: how is this any better than the same logic appearing in each transpiler?  e.g. transpiling as `a.contains(b)` in Rust is desirable to most use of `a in b`, not just `str` & `str`.  Me thinks this is taking the "rewriter" concept to far.

jayvdb: That said, there is a need for a generic rewriter, to convert `a.find(b) != -1` into `b in a` ;-)

py2many/rewriters.py
adsharma: Agreed and addressed in the updated commit.

py2many/rewriters.py
adsharma: Yeah - 4-5 per language rewriters would be cleaner. But this was a lot fewer lines of code, which is not always the right metric to optimize for.

I'm for getting this in first and breaking it up into N rewriters if this pattern becomes hard to maintain and sets a bad example.

jayvdb:
https://github.com/adsharma/py2many/blob/8ef607d/tests/test_unsupported.py#L66 should be fixed...?

py2many/rewriters.py
jayvdb: I dont get why it needs any rewriters.  I would put it in `visit_Compare`.

> if this pattern ... sets a bad example.

IMO using `_language` in this way does set a bad example.  Using `_language` for when one language is very odd seems reasonable, but when a rewriter needs special logic for various languages, that logic belongs under that directory.

Anyways, I am looking forward to it landing in any way you want to do it.  It will help us get to 90% coverage very quickly, which is a magical line in the sand for me.

adsharma: > I dont get why it needs any rewriters. I would put it in visit_Compare

For me it's the complexity. If we look at _visit_AssignOne() in pyrs - it's 110 lines. When we support all the corner cases and are transpiling many real world programs, I worry about how many special cases we have inlined.

Keeping them all in rewriters keeps the base case really simple. For example Assign could just be these lines:

https://github.com/adsharma/py2many/blob/main/pyrs/transpiler.py#L854-L862

everything else gets moved to many small functions.

Anyways, let me fix up the tests and get this in and then we can discuss how to make the code base more maintainable separately.

jayvdb:
Would be great if you could get https://github.com/adsharma/py2many/pull/265 in first

Issue #158, opened 2021-05-03
Add a directory mode


adsharma:
```
py2many.py --dir=/some/python/module <opts>
```

The idea is that all of the python files in the module are parsed and type inference works across file boundaries and we end up with module-py2many which contains transpiled code for the target language.

jayvdb:
We already have recursive mode, which does some of it, but not the  type inference across file boundary.

As part of this, we also need a project-init step for each language, that creates the necessary files to build the project for each language.  e.g. `Cargo.toml' and `lib.rs`.  Doing a project init for individual files is also beneficial as it mean we can remove the dependency on cargo-script, kscript, etc for running scripts, as we could then push deps into the appropriate meta files.

`cargo clippy` and `dart analyze` only work in a project.  (https://github.com/adsharma/py2many/pull/163 used `RUSTFLAGS` to workaround that limitation.  https://github.com/adsharma/py2many/issues/164 could have a limited solution for dart)

adsharma:
Yes - if I were to guess the most likely usage model, it seems a directory of python -> project dir in the target language.
I think we have all the setup in github actions yaml file. But more approachable user facing documentation would be nice too.

jayvdb:
Another task before getting serious about directory mode is module level attributes (variables, functions, classes) should be public visibility.

Dart and Julia and Kotlin attributes are public visibility by default.

I'm not sure about Nim - pointer https://github.com/nim-lang/Nim/pull/17706

- Rust done in https://github.com/adsharma/py2many/commit/1ecec7d1c74c50f6a8653a649aeb63482c0fb3bc , but globals are going to be a headache.

- Go (rewriting names): https://github.com/adsharma/py2many/issues/174

- C++ (extern?): https://github.com/adsharma/py2many/issues/175

jayvdb:
> I think we have all the setup in github actions yaml file.

We need to create and delete the project dynamically for each Python directory.
That project creation logic should be, IMO, part of the language definition instance, with the logic in each `py*` package.

adsharma:
Simple test case:

https://paste.ubuntu.com/p/PDXNnPj2dc/

adsharma:
Tests were added via #391

PR #269, opened 2021-05-24
CI: Lint generated Rust

jayvdb:


Cargo.toml
```diff
@@ -0,0 +1,140 @@
...
+[dependencies]
+flagset = "*"
+futures = "*"
+pyo3 = "*"
+structopt = "*"
+strum = "*"
+strum_macros = "*"
+lazy_static = "*"
+
+[lib]
```
adsharma, line 17: Any way to generate this dynamically, so when new tests are added, entries appear here magically?

adsharma (approved):

Cargo.toml
jayvdb: Ya, that is going to need to be done for https://github.com/adsharma/py2many/issues/158

Issue #277, opened 2021-05-25
pyrs: Declarations must appear before use


jayvdb:
The `NotImplementedError` occurring in https://github.com/adsharma/py2many/blob/28e3b38/tests/test_transpile_self.py#L32 is

```pt
self = <pyrs.transpiler.RustTranspiler object at 0x1085aa940>, node = <ast.Call object at 0x108c02ca0>, fname = 'NestingTransformer', fndef = <ast.ClassDef object at 0x108c02c10>

    def _visit_struct_literal(self, node, fname: str, fndef: ast.ClassDef):
        vargs = []  # visited args
        if not hasattr(fndef, "declarations"):
>           raise NotImplementedError(f"Missing declarations for {fname}")
E           NotImplementedError: Missing declarations for NestingTransformer

pyrs/transpiler.py:384: NotImplementedError
```

This is occurring whenever a function/method uses a class which is defined later in the same file.  In Python, all of the top level attributes are processed before any of the code within them, so logic in a function is not evaluated until all classes in a module are loaded.

pyflakes solves this by deferring visiting of the `def` (etc) until after the outer level has been visited.
e.g. https://github.com/PyCQA/pyflakes/blob/da197ee/pyflakes/checker.py#L917

That is complicated, but efficient.

The alternative is to do multiple passes.  We're already doing that, so doing a few more passes wont hurt.

A temporary solution is to restructure the logic in py2many/ so that the logic complies with the capabilities of `pyrs`.  I did this for one bit of `pyrs/` , but I didnt do it for `py2many/` in my PR https://github.com/adsharma/py2many/pull/259 because I felt it would involve too much code changes for code that I didnt write, re-effectively authoring those lines.  @adsharma, it would be great if you could do this, so we can get all of py2many to pass the trnspile phase, and then we can investigate how to get py2many to generate compile-able Rust code (that would do py2rs only probably) - self-hosting!

jayvdb:
Class declarations still need to appear before their use.

PR #282, opened 2021-05-25
py2many: handle funcs accessed out of order

adsharma:
Tested with:

```
from dataclasses import dataclass

def foo() -> "Point":
    return bar()

def bar() -> "Point":
    return Point(1, 2)

@dataclass
class Point:
    x: int
    y: int

if __name__ == '__main__':
    print(foo().x)
```

Fixes: https://github.com/adsharma/py2many/issues/277

### DIFF
--- a/py2many/clike.py
+++ b/py2many/clike.py
@@ -5,7 +5,7 @@ import sys
 
 from enum import IntEnum
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


adsharma:
The test above can't be checked in because generated code has other problems.

PR #285, opened 2021-05-25
transpile self: Add more languages

jayvdb:
Simplify a raise in cli.py to avoid tripping up
transpilers.

Raise `NotImplementedError` in py2many to avoid
suppressing `Exception`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -50,11 +50,15 @@ jobs:
           pip3 --version
           pip3 install setuptools tox
 
-      - name: Install Julia dependencies for ${{ runner.os }}.
+      - name: Install Julia packages
         shell: bash
         run: |
           julia -e 'using Pkg; Pkg.add("JuliaFormatter")'
           julia -e 'using Pkg; Pkg.add("SuperEnum")'
+
+      - name: Install dependencies for ${{ runner.os }}.
+        shell: bash
+        run: |
           if [ "$RUNNER_OS" == "Linux" ]; then
               sudo apt-get install --yes astyle
               curl -sSLO https://github.com/pinterest/ktlint/releases/download/0.41.0/ktlint
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,6 +33,8 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: '3.8'
+      - name: Set up Visual Studio shell
+        uses: egor-tensin/vs-shell@v2
 
       - name: Install go packages
         shell: bash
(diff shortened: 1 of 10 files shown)


.github/workflows/main.yml
```diff
@@ -50,11 +50,15 @@ jobs:
           pip3 --version
           pip3 install setuptools tox
 
-      - name: Install Julia dependencies for ${{ runner.os }}.
+      - name: Install Julia packages
```
adsharma, line 53: Ah makes sense. They were not all about Julia.

py2many/clike.py
```diff
@@ -225,13 +225,13 @@ def _typename_from_annotation(self, node, attr="annotation") -> str:
                 node.container_type = type_node.container_type
                 return self._visit_container_type(type_node.container_type)
             if typename is None:
-                raise Exception(f"Could not infer: {type_node}")
+                raise NotImplementedError(f"Could not infer: {type_node}")
```
adsharma, line 228: This one could be `TypeError` with `node.lineno`. I'm sure there is a format we can report the errors in that allows IDEs and other tools to implement "click to go to the error". But that can happen in the code that catches the exception.

jayvdb: It is the transpiler logic is not working properly, iirc this was Julia only which was triggering this for lots of pyjl/ py2many/ inputs.

Implementing a custom subclass of `NotImplementedError` which has extra attributes for context would be useful, but is beyond the planned scope of this PR.  When building that, the implementation should also consume that extra context information in a meaningful way in order to ensure it is designed correctly.

jayvdb: Julia Call in pyjl/transpiler.py and Go Return in pygo/transpiler.py trigger this.  In both cases `type_node` is "None".

I can see why you think `TypeError` is appropriate, but the logic in those transpilers is not "wrong" - it isn't using the wrong type when calling this function.  The calling logic is just not able to handle more complex scenarios.

The problem with raising normal python exceptions for any of these types of problems is then we need to mask them, and real uses of those exceptions, caused by bad python code, will also be masked, allowing regressions.

In the end, all of these `NotImplementedError` need to be eradicated.  In some cases, like here, they should be replaced with `assert ..` or even `TypeError`.

In addition we should create a special subclass of `NotImplementedError` for each scenario which are not "ever" going to be implemented, such as
- syntax which is not possible in the target language, 
- syntax we dont intend to support for any language, etc

jayvdb:
Windows coverage is now at 89% coverage.

jayvdb:
ping @adsharma , now green

tests/test_cli.py
```diff
@@ -198,46 +201,46 @@ def test_generated(self, case, lang):
...
+                self.assertFalse(proc.returncode)
+
+                if expect_failure:
+                    raise AssertionError(f"{case}{ext} passed unexpectedly")
 
         finally:
             if not self.KEEP_GENERATED:
                 case_output.unlink(missing_ok=True)
-            exe.unlink(missing_ok=True)
+                exe.unlink(missing_ok=True)
```
adsharma, line 243: So KEEP_GENERATED means keep the generated source as well as the binary.
Sounds good. Wanted to make sure this was intentional.

jayvdb: Ya, occasionally I need to run the exe afterwards to see why it is failing.

PR #288, opened 2021-05-26
Web based

adsharma:


adsharma:
http://transpyle.me/ has an experimental web UI
Not ready for prime time, but still a useful playground to gather feedback.

adsharma:
Doesn't look like clippy is taking the new argument from the github workflow. @jayvdb ?

jayvdb:
you need to add ` -A clippy::eq_op` to the other branch in clippy.yml

jayvdb:
If only `py2ma.ny` was possible...

pyrs/transpiler.py
```diff
@@ -120,15 +121,15 @@ def usings(self):
...
+        ) if not self._no_prologue else ""
         cargo_toml = f"""
         //! ```cargo
         //! [package]
         //! edition = "2018"
         //! [dependencies]
         //! {deps_str}
         //! ```
-        """
+        """ if not self._no_prologue else ""
```
jayvdb, line 132: IMO, this should go back to using `cargo-deps` which are shorter, and be preceeded with a one liner comment to run the script using `cargo-eval` (and we should switch to cargo-eval immediately if we go down this path).

`edition = "2018"` can be supplied as a CLI arg - it is a constant, and doesnt need to vary per script.

That way the reader can quickly see which libraries are being used, and can easily see how to invoke it.

py2many/cli.py
```diff
@@ -392,6 +397,9 @@ def main(args=None, env=os.environ):
     parser.add_argument(
         "--extension", type=bool, default=False, help="Build a python extension"
     )
+    parser.add_argument(
+        "--no-prologue", type=bool, default=False, help=""
```
jayvdb, line 401: IMO it would make sense to broaden the scope of this a bit, and also exclude the cpplint rule exception markup, and other type of superfluous annotations being added to the generated code.

adsharma:
> you need to add ` -A clippy::eq_op` to the other branch in clippy.yml

I have it here?
https://github.com/adsharma/py2many/blob/web_based/.github/workflows/clippy.yml#L39

> If only `py2ma.ny` was possible...

Why didn't I think of that one? I then discovered that transpyle is another python -> fortran transpyler with conference papers published!

pyrs/transpiler.py
adsharma: Agree that cargo-deps was shorter and preferable. I don't have a strong opinion on cargo script vs eval.
Whatever we choose, it'd be nice if it also works on rust playground. The web UI links to it now.

jayvdb:
The other branch: https://github.com/adsharma/py2many/blob/web_based/.github/workflows/clippy.yml#L35

Issue #97, opened 2021-04-22
External transpilers for additional languages


jayvdb:
Would you be interested in adding support for more languages by utilising existing py2x transpilers?  I'm especially interested in JS atm.

e.g. for JavaScript https://github.com/atsepkov/RapydScript does a good job of converting Python to JavaScript with a similar result as py2many provides. (unlike http://www.transcrypt.org/ which generates a bastard of the two languages)  I havent looked at whether https://github.com/kovidgoyal/rapydscript-ng is fundamentally the same still, but it appears to be.

https://github.com/pseudo-lang/pseudo-python also might be useful, even though it is a subset of Python, which excludes globals it seems (ugh), but it very likely has more of python supported than py2many in its current form. (dev has also stalled, it seems)

https://github.com/onelang/OneLang and https://github.com/jarble/transpiler/ look very interesting.

It would only involve an enhancement to cli.py , and a test module (which I think I would disable by default), which processed the test cases here against the external transpilers.

We could then build our own implementations side-by-side and swap over when the local impl is close to reaching parity with the external tool.

adsharma:
dart2js is another way to get there.

jayvdb:
Right.  The question is whether this project will use other transpilers if they have been pre-installed.  I am willing to do it, as it has slight benefits for me, but only if the PR is about how we do it, and not if it is within scope of this project.

adsharma:
I don't mind having one "external" transpiler, which takes a static configuration option (Essentially the command to run) that calls into another transpiler.

py2many.py --external=1 test/case/coverage.py

Any additional complexity worries me from a maintenance perspective. I'd like to keep the test matrix and test run times roughly the same as what we have now.

jayvdb:
> I don't mind having one "external" transpiler, which takes a static configuration option ..

This isnt enough to justify working on this enhancement.  If I wanted that, I would just invoke the transpiler directly.

To be useful, the extra languages needs to be in the cli.py settings, it knows how to invoke it, just like "native" languages.  That brings the "cheap win" of getting extra languages quickly, making it easier to for people to adopt py2many as opposed to using other tools.  The more languages the tool supports, the more likely a manager will approve a developer using py2many or allocating time to helping py2many.

> I'd like to keep the test matrix and test run times roughly the same as what we have now.

[...]

PR #290, opened 2021-05-26
WIP: Add JavaScript using dart2js

jayvdb:
Related to https://github.com/adsharma/py2many/issues/97


jayvdb:
Some of this has been merged already.  I no longer see any value in using dart2js for generating JS, but there are bits of this PR which I think may be useful for linting dart, and other bits useful in general.  I'll get back to this very soon.

jayvdb:
6e3017ffa44ce2b is likely still useful, but not worth keeping this PR open.

Issue #293, opened 2021-05-27
dict.get


jayvdb:
Python:
```py
def dict_get():
    CODES = {"KEY": 1}
    assert CODES.get("KEY") == 1


def dict_get_default():
    CODES = {"KEY": 1}
    assert CODES.get("MISSING", 2) == 2
```
Kotlin works with the former only, and Julia with the latter only.  All others fail.

Also we're going to run into nullable problems with `.get(x)` which returns None, e.g. in Dart which uses `a[x]` and returns nullable `String?`. In Dart we can avoid the nullable if there is a default with `map[x] ?? default`, but if default is not a constant then we need to track the variable to determine if is nullable or not.  Likely needs a nullable annotation like `lifetime`, but please lets try to disable "sound null" in dart and equivalent in other languages and not tackle nullability until there is a good sample of viable test cases to ensure it is sufficiently implemented that it isnt preventing progress.

Transpiles into Rust with errors:
```rust
error[E0605]: non-primitive cast: `Option<&i32>` as `i32`
  --> dict.rs:45:13
   |
45 |     assert!((CODES.get("KEY") as i32) == 1);
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^ an `as` expression can only be used to convert between primitive types or to coerce to a specific trait object

error[E0061]: this function takes 1 argument but 2 arguments were supplied
   --> dict.rs:50:20
    |
50  |     assert!((CODES.get("MISSING", 2) as i32) == 2);
    |                    ^^^ ---------  - supplied 2 arguments
    |                    |
    |                    expected 1 argument
    |
note: associated function defined here

error[E0605]: non-primitive cast: `Option<&i32>` as `i32`
  --> dict.rs:50:13
   |
50 |     assert!((CODES.get("MISSING", 2) as i32) == 2);
   |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ an `as` expression can only be used to convert between primitive types or to coerce to a specific trait object
```

adsharma:
Yeah - the main thing that's missing here is that transpiler doesn't understand the semantics of python's `dict.get()`.
I feel that at ~1000 lines of code $lang/transpiler.py is at around the limits of maintainability. Adding more code in there to make it understand `dict.get()` semantics (and other common python idioms) is a significant amount of complexity.

I like adding plugins.py to all languages (currently only rust has it) and handling it there. The key idea is to index any tables there with `dict.get` instead of strings.

PR #294, opened 2021-05-27
pydart dict fixes

jayvdb:
Relates to https://github.com/adsharma/py2many/issues/293


adsharma:
I'll merge everything till  acac608. Need more time for:

* understanding dart null safety. `a[key]!` seems more succinct and null safety is more in line with static compilation mindset of py2many.
* mutable state in the transpiler (_in_assign_node). For complex ASTs it could make it harder to reason about.

jayvdb:
Ya, I was going to suggest that.  Two notes:
- I got `a[key]!` working, except that `dart format` wouldnt accept it.
- `_in_assign_node` is horrid, but I couldnt think of a less complicated mechanism.  I am all ears if you have another pattern.  We should probably look at whether other languages also need similar decoration except when the expression is part of the LHS.  Very likely this occurs in other languages for `a[x]`, but there might be others also.

adsharma:
> Ya, I was going to suggest that. Two notes:
> 
>     * I got `a[key]!` working, except that `dart format` wouldnt accept it.

Is it because dart hasn't been migrated to handle null safety? Docs seem to suggest that this requires 2.12.

> 
>     * `_in_assign_node` is horrid, but I couldnt think of a less complicated mechanism.  I am all ears if you have another pattern.  We should probably look at whether other languages also need similar decoration except when the expression is part of the LHS.  Very likely this occurs in other languages for `a[x]`, but there might be others also.

The other pattern is the one in py2many/mutability_transformer.py, where the mutable state is isolated to a small piece of code (the transformer itself). So any bugs are easier to triage since it's isolated to that piece of code.

It's also possible to save the lvalue/rvalue info on the `ast.Subscript` node itself from a transformer, so the transpiler code is simplified.

adsharma:
Would it have the same behavior as python for cases such as:

```
#!/usr/bin/env python3

def main():
    a = [1, 2, 3, None, 5]
    for i in range(len(a)):
        print(a[i])
```

That is - can it distinguish between cases where the key was not found vs the key was found, but the value was null?
Also doesn't look like there was much progress on https://github.com/AmirKamali/Flutter_Vnum/issues/10

adsharma:
Also, I was able to do:

```
  dart pub upgrade --nullsafety collection lints sprintf tuple
```

and change dart version to 2.12 and use `a[i]!`. Seems to work. Let me ping that issue one more time to see if we can get a new release.

jayvdb:
The approach you are taking this would enforce that end-user projects must have null-safety enabled.

That isnt feasible for many projects, due to dependencies like our own problem with vnum.
See https://github.com/search?q=%22null-safety%22+language%3ADart+state%3Aopen&type=Issues for examples of the problem.  Including very important packages like https://github.com/dart-lang/intl_translation/issues/134 .

On the plus side, the issue mentioned above suggests a workaround which it might make it possible to override the dependency.

Unless you can see a better solution for v0.3.0, I think it is would best that we get this in, as it is good-enough to make Dart support more typical dicts.  And we can create an issue to improve it for the next release.

adsharma:
https://github.com/adsharma/py2many/pull/423 has a transformer to help with this, essentially move the `in_assign_node` logic out of line.

After that the commit will be small and easy to maintain.

I suspect there may be more languages which might want this behavior. Ok to keep it to dart for now.

If you choose to go the generic route, we'd rewrite the AST with `if a[i] is None: throw Exception(...)` and then transpile this using dart's `??` operator.

adsharma:
The bug in #423 has been fixed. Please try rebasing on top of it.

Also, having the stack ordered by pull request number makes it easier to review. That order would make this the lowest diff in the stack.

I have some concerns about the human readability impact of this on generated code. But on balance I think the impact is small (at least for the cases we have) and think that this diff should go in for greater python compatibility till dart ecosystem adopts null_safety more fully.

jayvdb:
Closing this.  I dont have the capacity to work on multiple PRs interleaved with yours atm.

Feel free to take over this PR in https://github.com/adsharma/py2many/pull/423, i.e. do the complete solution.  That will be more efficient.

PR #297, opened 2021-05-28
rewriter: ignored assign rewriter

adsharma:


adsharma:
```
(_c1, _, _c2) = (1, 2, 3)
```

transpiled to

```
{
 ...
  ({
    int _c1 = 1;
    (void)2;
    int _c2 = 3;
    ;
  });
}
```

passes tests, but is not quite right. If c1 and c2 are accessed further down in the parent block, they will be out of scope. We need to keep them in the same scope and not create a new one. That could be handled in a separate diff to keep this one more focused.

Issue #298, opened 2021-05-28
Assignment rewriter: handle scoping issues


adsharma:
Problem described here:

https://github.com/adsharma/py2many/pull/297

adsharma:
Fixed by the referenced commit

PR #301, opened 2021-05-29
Error propagation

adsharma:
Mostly for testing with_open.rs. The signature of main() is still being discussed elsewhere.

### DIFF
--- a/tests/cases/with_open.py
+++ b/tests/cases/with_open.py
@@ -1,10 +1,11 @@
 #!/usr/bin/env python3
 
+import os
 from tempfile import NamedTemporaryFile
 
 
 if __name__ == "__main__":
-    with NamedTemporaryFile(mode="a+") as temp_file:
+    with NamedTemporaryFile(mode="a+", delete=False) as temp_file:
         file_path = temp_file.name
         with open(file_path, "w") as f:
             f.write("hello")
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,9 @@ structopt = "*"
 strum = "*"
 strum_macros = "*"
 lazy_static = "*"
+pylib = "*"
+tempfile = "*"
+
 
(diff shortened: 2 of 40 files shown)


adsharma:
@jayvdb - do you know why windows is not able to create a temp file?

tests/cases/with_open.py
```diff
@@ -1,12 +1,14 @@
...
 
 if __name__ == "__main__":
-    with NamedTemporaryFile(mode="a+") as temp_file:
+    with NamedTemporaryFile(mode="a+", delete=False) as temp_file:
         file_path = temp_file.name
         with open(file_path, "w") as f:
             f.write("hello")
         with open(file_path, "r") as f:
             print(f.read())
+        os.unlink(file_path)
```
jayvdb, line 14: try moving this out of the outer `with`.

adsharma: Then `file_path` goes out of scope.  Looks like I have to make a more invasive change that also fixes https://github.com/adsharma/py2many/issues/298

jayvdb: It goes out of scope in Rust?

If that is it, wrap the unlink in a try block and ignore deleting the temporary file on Windows.  temp files are a special case, and windows is an even more special case ;-)

Or, instead of using temporary files for the test case, use a generated filename that is unique for each language, which can be supplied in the program args using [`CASE_ARGS`](https://github.com/adsharma/py2many/blob/main/tests/test_cli.py#L48), like `$TEMP/py2many-with-open-$lang.txt`.

adsharma: I did a temporary file with delete=False which is logically equivalent to generating a unique filename for each language.

adsharma:
Will go ahead and merge this for now. Let's revisit if the signature of main() needs to change.

PR #308, opened 2021-06-02
py2many: machine parseable error reporting

adsharma:
We could use this pattern anywhere we throw an error


adsharma:
Closed via 2261b5bb1e0ef0c81b3f94a3dac578abe5117514

Issue #240, opened 2021-05-21
C++ headers not found on GitHub Actions Ubuntu 20210517.1


jayvdb:
Last main commit which was working used

  Environment: ubuntu-20.04
  Version: 20210510.0
  Included Software: actions/virtual-environments@ubuntu20/20210510.0/images/linux/Ubuntu2004-README.md
  Image Release: actions/virtual-environments@ubuntu20%2F20210510.0 (release)

First broken commit was

  Environment: ubuntu-20.04
  Version: 20210517.1
  Included Software: actions/virtual-environments@ubuntu20/20210517.1/images/linux/Ubuntu2004-README.md
  Image Release: actions/virtual-environments@ubuntu20%2F20210517.1 (release)

https://github.com/actions/virtual-environments/blob/ubuntu20/20210510.0/images/linux/Ubuntu2004-README.md vs https://github.com/actions/virtual-environments/blob/ubuntu20/20210517.1/images/linux/Ubuntu2004-README.md has clang version changes, which are likely the cause.

jayvdb:
I've got a C++ PR ready, but need to solve this at the same time.

We probably also want to get ahead of the game and have CI checking multiple clang versions so that we can see this type of breaking coming down the pike.

jayvdb:
I've confirmed this is only happening on the latest Ubuntu image.
ubuntu-18.04 has python 3.6

adsharma:
Thanks for tracking this down! Is there someone responsible for the Ubuntu images on github looking into this?

jayvdb:
I couldnt find any issues in https://github.com/actions/virtual-environments/ that were related, so we should keep this issue here open.

adsharma:
https://github.com/actions/virtual-environments/issues/3376 seems to have some discussion that's related. My observation is that the errors aren't deterministic. And this issue seems to be indicating that it's reproducible on Ubuntu 18.04 too.

jayvdb:
Ya I was wondering about that one.  We can try their suggested workaround on ubuntu 20.  I havent seen any problems on Ubuntu 18

jayvdb:
There is a new pre-release https://github.com/actions/virtual-environments/releases/tag/ubuntu20%2F20210524.1
I havent tried it yet, as none of the clang related issues have been fixed.

PR #324, opened 2021-06-22
ci: try ubuntu 20.04 again

adsharma:
Fixes https://github.com/adsharma/py2many/issues/240


adsharma:
Conclusion: Doesn't work. Problem specific to 20.04 + libc++ (which is different from libstdc++)
Our best hope is to follow up here: https://github.com/actions/virtual-environments/issues/3494

Issue #229, opened 2021-05-15
Look into using pylib for nim


adsharma:
https://github.com/Yardanico/nimpylib

pynim/transpiler.py:visit_range(), visit_print() etc could be simplified using this library

adsharma:
Similar library for rust:

https://github.com/adsharma/pylib-rs

jayvdb:
IMO, use it, but have a fairly strict rule about when to use it, and when not to, and then break the rule as necessary to reach the goal quicker and leave a comment each time the rule is broken.

Something like: Only use helpers for Python functionality that is not feasible to reproduce using a few lines of the target language usings its stdlib.

i.e. we should inline routines where a person would normally inline routines, and use external libs where a person would normally do it, and err slightly on the side of inlining vs invoking external code, as surely one of the goals of py2many is to *show* people how to (correctly..?) rewrite Python logic into other languages.

jayvdb:
That library is still seeing a lot of development.

It looks like it will allow `with open(...)` and `with NamedTemporaryFile()` without any plugins, as it allows using Python names and (apparently) Python behaviour.

litlighilit:
Hello, while it's sad that the original url of pylib: <https://github.com/Yardanico/nimpylib> is 404...

yet nimpylib is still alive and even more active! new link here -> [github.com/nimpylib/pylib](https://github.com/nimpylib/pylib).

> That library is still seeing a lot of development

Now the newly born repo has implemented quite few more Python-like features, as well as some stdlibs.

litlighilit:
I'm coming when planning about [this issue](https://github.com/nimpylib/pylib/issues/9), and wonder if there would be some cooperation between two repos.  :)

jayvdb:
Thanks for advising us of the new project; great to see it is alive.

Unfortunately py2many is not so active at the moment.

However, I can review PRs and organise a new release of py2many if you are wanting to use it as a pre-processor of Python.

py2many has the ability for each language transpiler to add Python -> Python-like transformers, so doing `print(... , end=..)` -> `print(... , endl=..)` is trivial.

adsharma:
With mojo as a new entrant into this area, this idea of rewriting python stdlib from scratch without using C-API has taken a life of its own. Mojo seems to be doing it, not suprisingly in mojo. But the lack of a mojo -> python AST rewriter (I couldn't find an open source AST parser for mojo) makes it a bit harder.

[...]

PR #325, opened 2021-06-23
nim: migrate to plugin infra

adsharma:


jayvdb (approved):

jayvdb:
This closes https://github.com/adsharma/py2many/issues/229 ?

adsharma:
Doesn't close them yet, but gets us closer. 

The intent here is to use the generic dispatch code in py2many/clike.py, rather than the one in pynim (unfinished work from earlier PRs like https://github.com/adsharma/py2many/pull/236, which got too large to finish in one shot).

PR #327, opened 2021-06-23
pyrs: Handle indexing HashMap

jayvdb:
This includes a 'subset' of https://github.com/adsharma/py2many/pull/251 , and gets my project back onto the mainline of py2many.

### DIFF
--- a/pyrs/transpiler.py
+++ b/pyrs/transpiler.py
@@ -689,6 +689,12 @@ class RustTranspiler(CLikeTranspiler):
                 index_typename = get_inferred_rust_type(self._slice_value(node))
                 if index_typename != "u64" or index_typename != "usize":
                     index = self._cast(index, "usize")
+            is_dict = value_type is not None and value_type[0] == "Dict"
+            if is_dict:
+                value_type = getattr(node.value.annotation, "container_type", None)
+                index_typename = get_inferred_rust_type(self._slice_value(node))
+                if index_typename == value_type[1][0]:
+                    index = "&" + index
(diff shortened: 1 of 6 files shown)


pyrs/transpiler.py
```diff
@@ -224,6 +224,8 @@ def visit_FunctionDef(self, node, async_prefix=""):
                 else:
                     typename = self._generic_typename_from_type_node(node.returns)
                 typename = map_type(typename, extension=True, return_type=True)
+            if typename == "&str":
```
adsharma, line 227: How does it work for cases like this:

```
pub fn return_dict_index_int(key: i32) -> &'static str {
    let s = String::from("one");
    let slice: &str = &s;
    let CODES: &HashMap<i32, &str> = &[(1, slice)].iter().cloned().collect::<HashMap<_, _>>();
    return CODES[&key];
}
```

We have to have additional checks to make sure that the thing that's being returned has a static lifetime, which is true for simple test cases we have, but not true for the general case where we could be returning arbitrary strings.

I implemented enough lifetime analysis in inference.py to help with this case. Essentially any time we're propagating type information via node.annotation, we need to propagate node.annotation.lifetime for this to be complete.

https://paste.ubuntu.com/p/NGkYb4h2Pt/

After that you could write:

```
if node.returns.lifetime == Lifetime.STATIC:
   typename = f"&'static {typename}"
```   
   
Let me know if you want to incorporate it here or I'll send a separate pull req.

jayvdb: Thanks for doing that.

Issue #311, opened 2021-06-02
Question: tox testing target


PXshadow:
Hey I apologize if I'm missing something obvious. I've digged through the source code and can't seem to find any way to run the tests (preferably through tox) for only a single language (in my case golang). Thanks for any help on this, I'm very impressed thus far with the tooling, and speed of the compiler.

Also noob question, feel free not to answer but what are you supposed to do with all of the pregenerated coverage files?

adsharma:
```
tox -e py38 -- -k go 
```

should run only the golang tests.

Re: coverage files

They give us a clue about how much of the transpiler code remains untested after running tox. @jayvdb  is the expert on the topic.

jayvdb:
I am not sure what you mean by "pregenerated coverage files".

There is quite a lot of generated files emitted during the test invocation.  And there are `.coverage` files, which are in theory used by `pytest-cov`.

We havent been maintaining the `.gitignore` very well, but a lot of them should be listed there.

We also could change the test output directories so that it is easier to glob them in `.gitignore`.

PXshadow:
Yeah that would be helpful to glob more things in ``.gitignore`` , thank you for the command, it is working quite well, I'll close the issue.

PR #328, opened 2021-06-23
tests: Move builds into dedicated directory

jayvdb:
This problem was raised in https://github.com/adsharma/py2many/issues/311#issuecomment-853484870

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -82,8 +82,8 @@ jobs:
         shell: bash
         run: |
           dart --version
-          dart pub global activate stagehand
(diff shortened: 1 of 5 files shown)


jayvdb:
I think I need to move the generated output into `tests/build` for dart to pick up the generated `pubspec.yaml`.

adsharma:
tests/__init__.py seems to be missing. Could you add it in one of your other PRs?

jayvdb:
`tests/__init__.py` is only needed in py2

adsharma:
```
pytest-3 -k rust -v
```

fails without it.

PR #329, opened 2021-06-23
tests: Replace golint with revive 

### LINKED ISSUE #191: Reduce golint min_confidence to default 0.8
The current golint level of 1.0 causes several test cases to become errors, mostly due to capitalisation and docstrings.

However the default of 0.8 causes many more errors.

jayvdb:
Closes #191

### DIFF
--- a/revive.toml
+++ b/revive.toml
@@ -1,0 +1,30 @@
+ignoreGeneratedHeader = false
+severity = "warning"
+confidence = 0.8  # 0.1 also works
+errorCode = 1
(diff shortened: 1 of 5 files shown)


revive.toml
```diff
@@ -0,0 +1,30 @@
+ignoreGeneratedHeader = false
```
adsharma, line 1: I think you meant to have this change in the next diff in the stack.

jayvdb: fixed

Issue #291, opened 2021-05-27
Not implemented constructs


jayvdb:
We've talked a bit about how to handle not-implemented Python constructs.

Raising NotImplementedError, and building specialised more meaningful subclasses, is the road we're currently going down.

The problem with that approach is that it halts the transpiler at the first unsupported construct, preventing someone from seeing that most of their python input is able to be transpiled, and only a section of it is not supported.

Instead of this, we might enclose an un-transpiled segment into an assertion in the target language which causes the problem to fail when that part of the code in the target language is used.  The assertion would show the (ast of the) original Python code which couldnt be transpiled.  Ideally it also contains some constant in the generated code so that the transpiler invoker (e.g. the CLI, or test rig) can detect and report to the caller how many un-transpiled segments exist in the generated code.  This will also increase the coverage, because the transpiler will keep working after the unsupported node.

The transpilers currently have `visit_unsupported_body` which looks like it might have been trying to do something like this.

Ideally both modes are possible, as a py2many developer usually wants to see a backtrace, but an end-user wants a best-effort output.

adsharma:
Yes - behavior could be controlled with a flag. The end user might also want correctness guarantees in the form of validation errors from the transpiler.

PR #334, opened 2021-06-25
Add AstNotImplementedError

jayvdb:
Related to https://github.com/adsharma/py2many/issues/291

### DIFF
--- a/py2many/cli.py
+++ b/py2many/cli.py
@@ -12,13 +12,14 @@ from subprocess import run
 from typing import Callable, List, Optional
 
 from .analysis import add_imports
-from .clike import CLikeTranspiler
-from .scope import add_scope_context
 from .annotation_transformer import add_annotation_flags
+from .clike import CLikeTranspiler
+from .context import add_variable_context, add_list_calls
+from .exceptions import AstNotImplementedError
+from .inference import infer_types
 from .mutability_transformer import detect_mutable_vars
 from .nesting_transformer import detect_nesting_levels
-from .context import add_variable_context, add_list_calls
-from .inference import infer_types
+from .scope import add_scope_context
 
 from pycpp.transpiler import CppTranspiler, CppListComparisonRewriter
 from pyrs.inference import infer_rust_types
@@ -394,7 +395,10 @@ def _process_dir(settings, source, outdir, env=None, _suppress_exceptions=True):
                 format_errors.append(path)
         except Exception as e:
             print(f"Error: Could not transpile: {path}")
-            print(f"Due to: {e.__class__.__name__} {e}")
+            if isinstance(e, AstNotImplementedError):
+                print(f"Due to {e.lineno}:{e.col_offset}: {e.__class__.__name__} {e}")
+            else:
+                print(f"Due to: {e.__class__.__name__} {e}")
             failures.append(path)
             if _suppress_exceptions:
                 if _suppress_exceptions is not True:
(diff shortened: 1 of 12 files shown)


py2many/rewriters.py
```diff
@@ -117,7 +117,10 @@ def capitalize_first(name):
 def camel_case(name):
     if "_" not in name:
         return name
-    return "".join(capitalize_first(part) for part in name.split("_"))
+    # Dont rewrite dunders
+    if name[:2] == "__" and name[-2:] == "__":
```
adsharma, line 121: nit: name.startswith("__") and endswith

pydart/transpiler.py
```diff
@@ -510,8 +510,9 @@ def visit_unsupported_body(self, name, body):
         return buf
 
     def visit_Raise(self, node):
-        exc = self.visit(node.exc)
-        return f"throw new {exc};"
+        if node.exc is not None:
+            return "throw new {};".format(self.visit(node.exc))
```
adsharma, line 514: For self transpilation and consistency, f-strings work better here.

py2many/cli.py
```diff
@@ -394,7 +395,10 @@ def _process_dir(settings, source, outdir, env=None, _suppress_exceptions=True):
                 format_errors.append(path)
         except Exception as e:
             print(f"Error: Could not transpile: {path}")
-            print(f"Due to: {e.__class__.__name__} {e}")
+            if isinstance(e, AstNotImplementedError):
```
adsharma, line 398:
```
 if hasattr(e, "lineno"): ...
 ```
 is more precise here. I can imagine we'll want to throw different types of exceptions from the transpiler

File: py2many/cli.py [hunk 4]
```diff
@@ -472,7 +476,7 @@ def main(args=None, env=os.environ):
                 import traceback
 
                 formatted_lines = traceback.format_exc().splitlines()
-                if hasattr(e, "lineno"):
+                if isinstance(e, AstNotImplementedError):
```
adsharma, line 479: Ditto.

py2many/inference.py
```diff
@@ -440,9 +441,9 @@ def visit_BinOp(self, node):
                 left_id is not None
                 and (left_id, type(node.op)) not in LEGAL_COMBINATIONS
             ):
-                err = TypeError(f"{left_id} {type(node.op)} {right_id}")
-                err.lineno, err.col_offset = node.lineno, node.col_offset
-                raise err
+                raise AstNotImplementedError(
```
adsharma, line 444: `TypeError` was more precise here. One more instance below.

py2many/exceptions.py
```diff
@@ -0,0 +1,8 @@
+import ast
+
+
+class AstNotImplementedError(NotImplementedError):
```
adsharma, line 4: Perhaps make this a base class for capturing the lineno and then have `TypeError` and `AstNotImplementedError` as subclasses.

adsharma (commented):
Certainly helpful towards getting more detailed error messages when self transpiling.

py2many/cli.py
jayvdb: It would need to be `if hasattr(e, "lineno") and hasattr(e, "col_offset"):` which is silly.  Those properties should be defined in an interface.

Currently the interface is `AstNotImplementedError`.  If we need others, we can create `exceptions.AstErrorBase` and `AstNotImplementedError` and others can subclass that.

pydart/transpiler.py
jayvdb: We will need to support `str.format` way before we have a working transpiled transpiler.

py2many/cli.py
adsharma: Sound good. Let's check `isinstance(AstErrorBase)` and have two subclasses `TypeError` and `AstNotImplementedError`.

adsharma: I'll merge this as-is to make forward progress. Let's work on the Exception hierarchy in stacked diffs.

PR #335, opened 2021-06-25
Julia fixes

jayvdb:
Improve wrapper around format.jl and fix Julia default type.

### DIFF
--- a/tests/test_transpile_self.py
+++ b/tests/test_transpile_self.py
@@ -214,10 +214,7 @@ class SelfTranspileTests(unittest.TestCase):
             _suppress_exceptions=suppress_exceptions,
         )
         if FileNotFoundError not in suppress_exceptions:
-            assert set(successful) == {
-                transpiler_module / "__init__.py",
-                transpiler_module / "plugins.py",
-            }
+            assert_only_init_successful(successful)
 
         successful, format_errors, failures = _process_dir(
             settings,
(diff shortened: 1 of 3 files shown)


tests/test_transpile_self.py
```diff
@@ -188,10 +188,7 @@ def test_julia_recursive(self):
             _suppress_exceptions=suppress_exceptions,
         )
         if FileNotFoundError not in suppress_exceptions:
-            assert set(successful) == {
-                transpiler_module / "__init__.py",
-                transpiler_module / "plugins.py",
-            }
+            assert set([i.name for i in successful]) == {"__init__.py"}
```
adsharma, line 191: nit: `{i.name for i in successful}` for set comprehension

tests/test_transpile_self.py
```diff
@@ -199,7 +199,7 @@ def test_cpp_recursive(self):
 
     def test_julia_recursive(self):
         settings = self.SETTINGS["julia"]
-        suppress_exceptions = False if SHOW_ERRORS else (NotImplementedError,)
+        suppress_exceptions = (False,) if SHOW_ERRORS else (NotImplementedError,)
```
adsharma, line 202: I wish we could be a bit stricter with typing `suppress_exceptions`. I'll merge this as-is for now. Let's clean up later.

PR #336, opened 2021-06-25
Implement _process_many: a vectorized interface to the transpiler

adsharma:


py2many/cli.py
```diff
@@ -315,6 +315,42 @@ def _relative_to_cwd(absolute_path):
     return pathlib.Path(os.path.relpath(absolute_path, CWD))
 
 
+def _format_one(settings, output_path, env=None):
+    restore_cwd = False
```
adsharma, line 329: @jayvdb the test failures are due to "format.jl/ktlint" not available. But why only on windows?

I suspect it has something to do with the `os.chdir` logic in this function. Perhaps it's better to write:

```
try: 
  ...
finally:
    os.chdir(restore_cwd)
```

adsharma: I thought about it some more. Earlier we had a try-catch around `_process_one()`. After this change, we have a try-catch only around `transpile_one()`. So any exceptions from `format_one()` is getting propagated and logged as a test failure.

I added a try-catch around `format_one()` to maintain the same behavior as the old code.

However, we need to look into why windows has significantly fewer test passes vs ubuntu. I suspect it's setup problems with formatters.

jayvdb: It is a problem with invoking the formatters on Windows.  Its got to do with .bat wrappers, `COMSPEC`, and oddities about win32 exec.  I've been down that rabbit hole once, and didnt get close enough to a neat solution to get a PR up.

py2many/cli.py
```diff
@@ -73,16 +73,19 @@ def core_transformers(tree):
     return tree, infer_meta
 
 
-def transpile(filename, source, transpiler, rewriters, transformers, post_rewriters):
+def transpile(filenames, sources, transpiler, rewriters, transformers, post_rewriters):
```
jayvdb, line 76: IMO `_` prefix this, as its signature is likely to keep changing in backwards-incompatible ways.

py2many/cli.py
```diff
@@ -366,15 +360,89 @@ def _process_once(settings, filename, outdir, env=None):
...
             os.chdir(restore_cwd)
+    except Exception as e:
+        print(f"Error: Could not format: {output_path}")
+        print(f"Due to: {e.__class__.__name__} {e}")
+        return False
 
     return True
 
 
+def _get_output_path(filename, ext, outdir):
```
jayvdb, line 371: could you put this above `_format_one` so that its inner code is no longer being moved - git will just mark the lines around it as changed.

py2many/cli.py
```diff
@@ -366,15 +360,89 @@ def _process_once(settings, filename, outdir, env=None):
...
+    output_path = outdir / (filename.stem + ext)
+    if ext == ".kt" and output_path.is_absolute():
+        # KtLint does not support absolute path in globs
+        output_path = _relative_to_cwd(output_path)
+    return output_path
+
+
+def _process_once(settings, filename, outdir, env=None):
```
jayvdb, line 379: likewise putting this between `_get_output_path` and `_format_one` will result in a smaller diff

jayvdb (changes requested):
PR title "Directory mode" is misleading as it doesnt yet do anything new, like cross-module analysis, or initialising the directory for a different language (i.e. create Cargo.toml/pubspec.yaml).

py2many/cli.py
adsharma: I renamed to `_transpile_one` and `_process_one` and changed the signature of this function with types so hopefully it's more stable going forward.

jayvdb: It wont be stable :P

jayvdb (approved):

py2many/cli.py
adsharma: Renamed. I think you're trying to say `main()` is the only stable interface.

jayvdb: ya, I expect all of the guts of `cli`, except for `main`, will be moved to different modules, and become more structured.

This PR is the perfect opportunity to make it private, as it was already fiddling with all of the relevant uses of the function.

py2many/cli.py
```diff
@@ -386,26 +451,10 @@ def _process_dir(settings, source, outdir, env=None, _suppress_exceptions=True):
...
-                print(f"Due to: {e.__class__.__name__} {e}")
-            failures.append(path)
-            if _suppress_exceptions:
-                if _suppress_exceptions is not True:
-                    if not isinstance(e, _suppress_exceptions):
-                        raise e
-            else:
-                raise e
+    successful = _process_many(settings, input_paths, outdir, env=env)
+    failures = set(input_paths) - set(successful)
```
jayvdb, line 457: this is "wrong".

failures has previously meant failed to transpile.  Not failed to format.

`_process_many` needs to return `format_errors` and `successful`

jayvdb (changes requested):

py2many/cli.py
jayvdb: https://github.com/adsharma/py2many/pull/335 has extra assertions which verify that `successful` is being populated correctly.

adsharma: Yes - I thought about it, but felt we could defer it until we figure out how to invoke the formatters for the _process_many() case.

Ideally, I'd like slow formatters like ktlint and format.jl to be invoked once on N files for good user experience. But I'm not sure if we have enough error parsing logic to figure out which files had format errors.

PR #342, opened 2021-06-26
py2many: cross module type inference

adsharma:
Passes the test case below. bar1() defined in another
file is correctly inferred as returning an int.

Tested via: py2many --rust=1 /tmp/testdir

containing the two files in the paste attached to the issue

Related: https://github.com/adsharma/py2many/issues/158

### DIFF
--- a/py2many/clike.py
+++ b/py2many/clike.py
@@ -441,7 +441,9 @@ class CLikeTranspiler(ast.NodeVisitor):
                 buf.append("if({0}) {{".format(self.visit(node.test)))
             else:
                 buf.append("if {0} {{".format(self.visit(node.test)))
-        buf.extend([self.visit(child) for child in node.body])
+        body = [self.visit(child) for child in node.body]
+        body = [b for b in body if b is not None]
+        buf.extend(body)
 
         orelse = [self.visit(child) for child in node.orelse]
         if orelse:
@@ -513,7 +515,7 @@ class CLikeTranspiler(ast.NodeVisitor):
         return (target, type_str, val)
 
     def visit_Delete(self, node):
-        raise AstNotImplementedError("del not implemented", node)
+        return "Del /* Unimplemented */"
 
     def visit_ClassDef(self, node):
         bases = [get_id(base) for base in node.bases]
--- a/py2many/clike.py
+++ b/py2many/clike.py
@@ -133,6 +133,7 @@ class CLikeTranspiler(ast.NodeVisitor):
         self._func_usings_map = {}
         self._attr_dispatch_table = {}
         self._keywords = {}
+        self._throw_on_unimplemented = True
 
     def headers(self, meta=None):
         return ""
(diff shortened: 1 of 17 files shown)


adsharma:
Triggers some extra failures in self transpilation, because now errors in imported files are attributed to the importing file as well.
Looking into the errors to figure out the best strategy to deal with the failures.

adsharma:
The main remaining problem to solve is:

```
pyrs/clike.py: AttributeError: 'Module' object has no attribute 'scopes'
```

which shows up during transpilation

py2many/clike.py
```diff
@@ -513,7 +515,7 @@ def visit_AnnAssign(self, node):
         return (target, type_str, val)
 
     def visit_Delete(self, node):
-        raise AstNotImplementedError("del not implemented", node)
+        return "Del /* Unimplemented */"
```
jayvdb, line 518: why? this would be a regression towards the goal of https://github.com/adsharma/py2many/issues/291

pycpp/transpiler.py
```diff
@@ -652,3 +655,18 @@ def visit_Print(self, node):
             else:
                 buf.append("std::cout << {0} << std::endl;".format(value))
         return "\n".join(buf)
+
+    def visit_DictComp(self, node):
+        return "DictComp /*unimplemented()*/"
```
jayvdb, line 660: these should all be an `AstNotImplementedError`.  Converting it to a string, or an assertion,  per https://github.com/adsharma/py2many/issues/291 , needs to occur at levels where that can be done sanely.

jayvdb (changes requested):

py2many/clike.py
adsharma: I think you want `visit_unsupported_body()` which results in a runtime assertion failure in the target language.
Once we raise AstNotImplementedError(), all subsequent code is not transpiled.

We could have a flag that controls whether `visit_unsupported_body()` throws or generates runtime assertion failure.

adsharma: Also we have only two instances of these:

```
del target.elts[i]
```

This could transpile to the following in cpp:

```
target_elts.erase(target_elts.begin() + i)
```

jayvdb: No.  The unsupported node raises an exception at that node.  Nodes which can represent unsupported nodes then catch that exception and generate an alternative representation if the user asked for it.  The nodes which can easily do this are functions, classes, etc.  Assignment nodes will eventually be able to do it right.

adsharma: That's different from what you wrote in #291. The problem that this code is trying to fix is the following pattern:

```
if condition:
   body-with-unsupported
else:
  body-with-supported
```

Then:

```
def visit_If(...):
      ...
      return "\n".join(buf)
```

The unsupported-body transpiles to `None` (because `visit_$Construct` was missing) and then when we try to join it, things break.

I'm not convinced that raising an exception for unsupported is the right answer for all cases. For example, even if you catch it somewhere else, you'll never visit the `else` part of the expression above won't uncover further transpilation problems. It's not clear that it's 1 or 2 places where the exception needs to be caught.

I don't want to hold this work while we explore a solution to that problem. Let's use #291 to discuss it. Returning a string fixes the problem with join() and allows us to make forward progress. We can make the behavior controlled by a flag, return a different string etc if it makes sense.

adsharma:
It was a bit non obvious to figure out why windows was failing. Turns out that windows runs both py38 and py39. 
With py38 test_cpp_recursive has 11 successful
With py39 it has 12 successful (probably due to ast changes)

I'll change the check to > 11 so they both pass.

PR #344, opened 2021-06-27
Fix and refine test_error_cases

jayvdb:
This also shows why the transpilers should not use TypeError.  As predicted, that causes real exceptions to not be noticed.

### DIFF
--- a/tests/test_unsupported.py
+++ b/tests/test_unsupported.py
@@ -12,6 +12,7 @@ import astpretty
 from unittest_expander import foreach, expand
 
 from py2many.cli import _get_all_settings, _relative_to_cwd, _transpile, _transpile_one
+from py2many.exceptions import AstIncompatibleAssign
 
 import py2many.cli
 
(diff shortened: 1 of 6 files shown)


tests/test_unsupported.py
```diff
@@ -308,26 +309,14 @@ def test_error_cases(self, case):
...
-        try:
-            _transpile([pathlib.Path("stdin.py")], [tree], settings,)[
-                0
-            ][0]
-        except Exception as e:
-            assert type(e) == expected_error
+        source_data = "from ctypes import c_int8 as i8, c_int16 as i16\n" + case
+        tree = get_tree(source_data, ext)
+        # This is supposed to raise a TypeError, however it doesn't.
+        _transpile(
```
adsharma, line 315: _transpile() is eating the exception. Perhaps call _transpile_one()?
I tested that the TypeError is being produced when run via py2many.py --rust=1 /tmp/foo.py

jayvdb: Will do.

jayvdb: done

Issue #319, opened 2021-06-16
V language - join efforts with py2v


dumblob:
py2many looks like an interesting endeavour (I've just discovered it through https://news.ycombinator.com/item?id=27165366 ).

I wonder whether there would be some benefits from sharing some knowledge, experience, abstractions, code, templates, etc. with https://github.com/vlang/py2v .

adsharma:
Looks interesting. I can see potential collaboration in a few areas:

* API plugins - translating the python stdlib to whatever V is using. py2many/py*/plugins.py (rust, cpp have it working. Other languages need more work)
* Some type of a pattern matching abstraction for python. Does V use match as a statement or an expression?
* Use of rewriters saved us from an explosion in complexity

I hope to make some documentation/video about the structure of py2many at some point. That could lead to a more in-depth discussion of these topics.

dumblob:
@div72 @medvednikov thoughts?

div72:
As of now, I don't think that py2many can benefit from py2v while py2v can benefit from py2many. Even if py2many had a V backend(which nsauzede has a simple PoC on their fork of py2many that I made some small improvements locally) since py2v is mostly written in V sharing work between them wouldn't be trivial.

That said I am planning on making a type analyzer which populates python ast with type annotations in the future. I believe that could be useful for all backends of py2many.

adsharma:
Thank you for sharing your plans @div72. Does the type analyzer you're thinking about work differently from the one py2many has?

https://github.com/adsharma/py2many/blob/main/py2many/inference.py

Inferred types are stored in an attribute called "annotation" in the AST node.

div72:
Yes, it'll also use annotation fields to store the types, although the implementation will differ.

adsharma:
Are you planning to do a Hindley–Milner type inference or something else?

The current py2many implementation follows the rules of that type system, but has implementation holes such as:

```
def foo(a: List[T]) -> T: ..

x = foo([1, 2, 3])
```
We should infer the type of x to be int, but the code doesn't.

https://github.com/adsharma/py2many/issues/246 is also related.

div72:
Planning on using abstract interpretation.

Seems like py2many's type inference also needs some work.

adsharma:
> abstract interpretation

We use eval() to handle cases where someone does "import foo as bar". By using type classes instead of strings in our type maps, the translation is more robust.

https://github.com/adsharma/py2many/blob/main/py2many/clike.py#L95

[...]

PR #350, opened 2021-06-27
Implement target language Python 

jayvdb:
This ensures that the re-writing results in valid Python
so that new languages never need to parse AST which has
invalid Python injected into it.

Also it provides an annotated intermediate format that
other tools can ingest to benefit from the py2many inference.

Related to https://github.com/adsharma/py2many/issues/319


py2many/python_transformer.py
```diff
@@ -0,0 +1,20 @@
+import ast
+
+from .clike import CLikeTranspiler
+
+
+class PythonTranspiler(CLikeTranspiler):
+    NAME = "python"
+
+    def visit(self, tree):
+        return ast.unparse(tree) + "\nif __name__ == '__main__': main(0, sys.argv)"
```
jayvdb, line 10: this bit needs to be rewritten to use https://github.com/adsharma/py2many/pull/349

`ast.unparse` is py3.9+, but I expect to be able to find a good backport that can be used to support py3.8

jayvdb:
The reason I have created a separate module for the Python "transpiler" is that it will provide a place to move some of the `cli` logic, and so that we can more clearly split py2many into two distinct phases:-

1. rewriting the input Python to be a more digestable subset of Python, with annotations where they can be injected into the Python
2. transpilng that Python to the target language

Keeping them distinct ensures that the first can become a more stable interface, where there is no use of `_language` or other tricks, so that it can be separately tested and used, e.g. by other transpilation tools.  The logic/output of phase 1 can be further ahead of any of the transpilers, e.g. cross-module inference atm, and the generated python should be usable to validate the first phase, e.g. be fed back into the test suite for the input Python, or pushed through mypy and other Python inference tools to ensure our inference isnt missing something that other tools pick up.

At the moment the new module only exposes a node visitor which emits a string, however it should also expose a tree transformer so that there is no extra penalty for having two phases (i.e. parse-visit-unparse-reparse-visit..).

adsharma:
Yeah - just taking advantage of the type inference is value add for some use cases.

I think a major reason many will be hesitant to use this is because AST loses comments and people don't want their comments to be nuked.

https://libcst.readthedocs.io/en/latest/why_libcst.html

But it doesn't seem to be API compatible with python's AST. It'd be great if you can estimate the amount of work needed to bridge the gap between libCST and ast.

py2many/python_transformer.py
```diff
@@ -0,0 +1,28 @@
...
+class PythonTranspiler(CLikeTranspiler):
+    NAME = "python"
+
+    def visit(self, tree):
+        return unparse(tree) + "\nif __name__ == '__main__': main()"
+
+    def usings(self):
+        return "\n".join(
+            [
+                "from typing import Callable, Dict, List, Set, Optional",
```
jayvdb, line 23: It would be useful to post-process with `autoflake` or https://github.com/hakancelik96/unimport to remove unused imports.

tests/test_cli.py
```diff
@@ -110,6 +111,10 @@ def get_python_case_output(case_filename, main_args, exit_code):
     return proc.stdout
 
 
+def standardise_python(code):
```
adsharma, line 114: A comment on what this is doing would help

py2many/rewriters.py
```diff
@@ -302,6 +302,10 @@ def visit_Compare(self, node):
                     ret = ast.parse("a.find(b) != string.npos").body[0].value
                     ret.left.func.value = right
                     ret.left.args[0] = left
+                elif self._language == "python":
```
adsharma, line 305: Why not short circuit on line 283?

adsharma (changes requested):
Looks pretty close. A few minor issues.

The command line usage needs change:

```
py2many.py --python=1 /tmp/foo.py
```

overwrites original source. Can we change the default to not overwrite and implement a flag?

```
a = 100
```

should be rewritten to:

```
a: int = 100
```

To make it happen, we need a rewriter that notices:

targets[0].annotation and converts `Assign` to `AnnAssign`.

py2many/rewriters.py
jayvdb: I was thinking that it might be better for the "simplified" Python to be using `str.find`, as that is easier to detect, and is able to be handled with the plugins system.  In many languages there is an equivalent function which returns `-1`.

Anyway, a short circuit on line 283 is better for now.

jayvdb:
> Can we change the default to not overwrite and implement a flag?

Created https://github.com/adsharma/py2many/issues/355

> .. notices `targets[0].annotation` and converts `Assign` to `AnnAssign`.

I'll do that now as it is pretty fundamental to what a Python target is trying to achieve.

PR #352, opened 2021-06-28
Finish moving dispatch to plugins

jayvdb:
Dart, Go & Kotlin.

### DIFF
--- a/py2many/rewriters.py
+++ b/py2many/rewriters.py
@@ -129,7 +129,8 @@ def rename(scope, old_name, new_name):
 
 
 class PythonMainRewriter(ast.NodeTransformer):
-    def __init__(self, language):
+    def __init__(self, main_signature_arg_names):
+        self.main_signature_arg_names = set(main_signature_arg_names)
         super().__init__()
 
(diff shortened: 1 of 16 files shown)


py2many/rewriters.py
```diff
@@ -145,10 +146,17 @@ def visit_If(self, node):
             if hasattr(node, "scopes") and len(node.scopes) > 1:
                 rename(node.scopes[-2], "main", "main_func")
             # ast.parse produces a Module object that needs to be destructured
-            ret = ast.parse("def main(argc: int, argv: List[str]): True").body[0]
+            if self.main_signature_arg_names == {"argc", "argv"}:
```
adsharma, line 149: I like this solution!

jayvdb: (wrong PR)

btw, if you give me non-main branch access, I can create PRs which use a prior PR as a base, so that each PR doesnt include preliminary commits. c.f. https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request

PR #357, opened 2021-06-29
v: initial v support

div72:


nsauzede:
I just tested your PR it works real fine @div72 thanks !
it can be merged upstream if @jayvdb @adsharma

jayvdb:
Several of the test cases fail with exception in the transpiler: `'VTranspiler' object has no attribute 'IGNORED_MODULE_LIST'`.  Note sure if this might be wanting something like `py2many.analysis.IGNORED_MODULE_SET`

sys_exit.v fails with "sys_exit.v:1:5: error: unexpected name `sys`".  This might be related to the ignored modules list.  I dont see a `sys` in https://github.com/vlang/v/tree/master/vlib

It would be nice if we could get test case `print` to work, as that is a pre-cursor for many of the other test cases to work.  Currently it fails with an error which looks like it might be easy enough to solve:
```
print.v:12:1: error: unexpected `else`, expecting `name`
   10 | if c:
   11 |     "True"
   12 | else:
      | ~~~~
   13 |     "False"
   14 | }')
```

pyv/transpiler.py
```diff
@@ -0,0 +1,612 @@
...
+        return f"@[{elements}]"
+
+    def visit_Set(self, node):
+        self._usings.add("sets")
+        elements = [self.visit(e) for e in node.elts]
+        elements_str = ", ".join(elements)
+        return f"toHashSet([{elements_str}])"
+
+    def visit_Dict(self, node):
+        self._usings.add("tables")
```
jayvdb, line 408: `dict.v` fails with
```
dict.v:1:8: warning: module 'tables' is imported but never used
    1 | import tables
      |        ~~~~~~
    2 |
    3 | fn implicit_keys() bool {
builder error: cannot import module "tables" (not found)
```

Is that a Vlang module which needs to be installed into CI?

Or is this copied from pynim?  It seems a few of these were copied from there.

Syntax which isnt supported yet can raise [`AstNotImplementedError`] found in https://github.com/adsharma/py2many/blob/main/py2many/exceptions.py

nsauzede:
TBH, you should consider my original V branch as highly experimental;
Yes I did copy/inspire from nim/go/rust (can't remember -- probably a mix), until I had something that transpiled ok
eg: this python tranpiles fine to V:
```python
def main():

    print('Hello, World!')

# Python3 program to add two numbers 

    num1 = 15

    num2 = 12

# Adding two nos 

    sum = num1 + num2



# printing values 

    print("Sum of ", num1, " and ", num2, " is ", sum) 



if __name__ == '__main__':

    main()
```
V output:
```v
fn main_func() {

	println('${'Hello, World!'}')

	num1 := 15

	num2 := 12

	sum := (num1 + num2)

	println('${'Sum of '} $num1 ${' and '} $num2 ${' is '} $sum')

}



fn main() {

	main_func()

}
```
If there are strange non-V generated imports, then it must be due because I didn't bother/failed to strip those from the existing languages I based upon (nim/v/rust etc..)

pyv/inference.py
```diff
@@ -0,0 +1,154 @@
+import ast
+
+from py2many.inference import get_inferred_type, InferTypesTransformer
+from py2many.analysis import get_id
+
+
+V_TYPE_MAP = {
+    "int": "int",
```
dumblob, line 8: Should be BigInt to be correct IMHO.

div72: While it'd be more correct, the performance impact would be a lot. Not a lot people go to values higher than i64.

dumblob: Then it should at least be `i64`, right? V's default `int` is `i32` if I'm not mistaken.

pyv/transpiler.py
```diff
@@ -0,0 +1,593 @@
...
+        fields = "\n".join([self.indent(f) for f in fields])
+        return f"type {node.name} = enum\n{fields}\n\n"
+
+    def visit_List(self, node):
+        elements = [self.visit(e) for e in node.elts]
+        elements = ", ".join(elements)
+        return f"[{elements}]"
+
+    def visit_Set(self, node):
+        raise AstNotImplementedError("Sets are not implemented in V yet.")
```
jayvdb, line 395: While generating `global2.v`: `global2.py:10:6: py2many.exceptions.AstNotImplementedError: __init__() missing 1 required positional argument: 'node'`.

Pass `node` as second arg to `AstNotImplementedError`.

jayvdb:
As V doesnt implement sets, you could make the following change
```diff
diff --git a/tests/cases/coverage.py b/tests/cases/coverage.py
index 936d618..0802af9 100755
--- a/tests/cases/coverage.py
+++ b/tests/cases/coverage.py
@@ -57,8 +57,6 @@ def show():
     print(len(a5))
     a9: List[str] = ["a", "b", "c", "d"]
     print(len(a9))
-    a6 = {1, 2, 3, 4}
-    print(len(a6))
     a7 = {"a": 1, "b": 2}
     print(len(a7))
     a8 = True
```
Then coverage.v fails due to
```
coverage.v:13:12: warning: use `x := []Type{}` instead of `x := []Type`
   11 | fn indexing() int {
   12 |   mut sum := 0
   13 |   mut a := []int([])
      |            ~~~~~~
   14 |   for i in (0..10 - 1) {
   15 |     a <<(i)
coverage.v:13:17: error: expression evaluated but not used
   11 | fn indexing() int {
   12 |   mut sum := 0
   13 |   mut a := []int([])
      |                 ~~~~
   14 |   for i in (0..10 - 1) {
   15 |     a <<(i)
```
Is that easy to fix?  It would be great if test case `coverage` passes, as that is our broad but basic test case.

If you want sys_argv to work, which complements `sys_exit`, the transpiler needs to declare whether it needs argc and/or argv, like https://github.com/adsharma/py2many/blob/d5ef9c3/pycpp/transpiler.py#L134 , and then in `visit_FunctionDef` it can fiddle with the types of those args like https://github.com/adsharma/py2many/blob/d5ef9c3/pycpp/transpiler.py#L189

div72:
A few more test cases(including coverage) should work after vlang/v#10619 is merged.

pyv/inference.py
div72: Probably a V bug but you can't `i64(1) in [1, 2, 3]` currently which breaks a test case. I will leave it as it is for now.

pyv/transpiler.py
```diff
@@ -0,0 +1,486 @@
...
+
+    def _combine_value_index(self, value_type, index_type) -> str:
+        return f"{value_type}{index_type}"
+
+    def comment(self, text):
+        return f"// {text}\n"
+
+    def _import(self, name: str) -> str:
+        if name == "sys":
+            return "import os"
```
jayvdb, line 72: i guess this should be using `self._usings`

pyv/transpiler.py
```diff
@@ -0,0 +1,486 @@
...
+        )
+        buf.append("}")
+        return "\n".join(buf)
+
+    def visit_Str(self, node):
+        return super().visit_Str(node)
+
+    def visit_Bytes(self, node):
+        raise AstNotImplementedError("Bytes are not supported yet.", node)
+        bytes_str = "{0}".format(node.s)
```
jayvdb, line 277: dead code

pyv/transpiler.py
```diff
@@ -0,0 +1,486 @@
...
+
+    def visit_cast(self, node, vargs) -> Optional[str]:
+        typ = get_id(node.func)
+        if typ == 'bool':
+            if get_inferred_v_type(node.args[0]) in V_WIDTH_RANK:
+                return f'{self.visit(node.args[0])} != 0'
+            else:
+                return f'({self.visit(node.args[0])}).bool()'
+
+    def _dispatch(self, node, fname: str, vargs: List[str]) -> Optional[str]:
```
jayvdb, line 183: As of https://github.com/adsharma/py2many/pull/352, all the other transpilers now use a generic `_dispatch` from the super class, which relies on bits of this function and `visit_print`/`visit_range`,etc being moved into `pyX/plugins.py`

pyv/inference.py
adsharma: We chose to go with `int` or `i32` in other backends.

The reasoning is that we can't obviously provide 100% compatibility with python's arbitrary precision `int` without making serious compromises on efficiency and still deliver on the promise of being able to write secure integer code without overflows/underflow errors using fixed width ints.

I like to think about the subset of python supported by py2many as a new statically typed programming language with the semantics of the language we're transpiling to, but with the syntax of python (so we can leverage the ecosystem to the extent that we can).

adsharma:
@div72 are you looking into the lint and test errors?

I can write up some docs about the development flow if that makes it easier.

div72:
> I can write up some docs about the development flow if that makes it easier.
@adsharma Thank you, I've been following @jayvdb comments but I think it would be nice to have something like CONTRIBUTING.MD for everyone.

> are you looking into the lint and test errors?
I fixed the lint but I am not exactly sure on why the build CI is failing.

adsharma:
> I fixed the lint but I am not exactly sure on why the build CI is failing.

coverage.py is failing for languages other than vlang, because you commented out two lines, but didn't update the "expected out".
You can run:

```
export UPDATE_EXPECTED=1
pytest-3 -k coverage -v
```

and add the diff to the same commit.

There are also a bunch of vlang specific test failures.  A quick way to see them all is:

```
pytest-3 -k vlang -v
```

adsharma:
#394 has some more documentation.

div72:
Do you all have every language and their formatters/packages installed or is there something I'm missing like a docker container?

adsharma:
> Do you all have every language and their formatters/packages installed or is there something I'm missing like a docker container?

I do it manually. But it should be possible to setup a docker ubuntu-latest image with the setup actions in `main.yml` @jayvdb ?

.github/workflows/main.yml
```diff
@@ -37,6 +37,17 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: '3.8'
+      - name: Install V
```
adsharma, line 40: Any chance of making binary distributions? I'll watch the amount of time it adds to CI.

pyv/plugins.py
```diff
@@ -0,0 +1,73 @@
...
+SMALL_USINGS_MAP = {}
+
+DISPATCH_MAP = {}
+
+MODULE_DISPATCH_TABLE = {}
+
+DECORATOR_DISPATCH_TABLE = {}
+
+CLASS_DISPATCH_TABLE = {}
+
```
adsharma, line 64: pynim/plugins.py has some type annotations. Perhaps add them here as well?

.github/workflows/main.yml
adsharma: https://github.com/adsharma/py2many/pull/390/checks?check_run_id=3013635796 took 24 mins and this commit took 27 mins. Not sure what the run-to-run variation is, but 3 mins extra isn't the end of the world. But these things can add up over time.

adsharma:
This looks very close. Any chance you could look into the two failures on mac? They seem to be related to dict keys/values.

.github/workflows/main.yml
div72: The runtime variation is more likely caused by GitHub itself, as https://github.com/adsharma/py2many/pull/390/checks?check_run_id=3015808449 also took 27m.

I don't like using the weekly binary distributions as since V is still evolving at a fast rate, they tend to lag. One of V's selling points is fast compilation and I don't really think it's much of an issue, although I can still use the weekly builds if you want.

adsharma: Takes 11 secs + clone time on my machine. So ok for now.

adsharma:
https://paste.ubuntu.com/p/bFnZ2myCNh/ could get the tests to pass. Still not clear why the test passes on windows/ubuntu, but fails on mac.

div72:
Apologies for the delay, I have been a bit busy. Will hopefully fix it tomorrow.

adsharma:
@div72 my suggestion above won't work because then the test is expected to fail on ubuntu, but it passes. So the problem boils down to this snippet:

```
1 in map{
	1: 1
}.keys().map(map{
	1: 1
}[it])
```

It passes on ubuntu, but fails on macos. I don't have a machine handy to debug, but hopefully the isolated test case is enough for you to take care of it upstream.

div72:
@adsharma I made an issue in V. In this case, macOS is the one right. The generated code should be invalid but as tcc permits it, both Windows and Linux CIs pass. Using temps should fix the issue, I'll probably do that tomorrow rather than waiting for a fix on V's side to not delay this further.

adsharma:
* Rebase against a recent main
* To fix the test_directory error:

```
pytest-3 -k "test_dir and vlang" -v
cp -a tests/dir_cases/test1-vlang-generated tests/dir_cases/test1-vlang-expected
git add tests/dir_cases/test1-vlang-expected/*
```
* The failure on binit looks like a formatter error

div72:
@adsharma I rebased locally, fixed the binit issue(needed to apply c7419cfc6c66df88bdb9d70d49d77bdc99dddd2e). However the dir case is an issue. V does not allow file imports, since the files are under the same directory they must have the same module name. Is there any way to tell if a dir is being processed and fetch it's name?

adsharma:
Yes, the output_dir is passed from here:

https://github.com/adsharma/py2many/blob/main/tests/test_cli.py#L441

You can use pathlib.Path.is_directory()

div72:
@adsharma I am talking about the transpiler stage, need to generate correct `module <name>` statements for all files.

adsharma:
> @adsharma I am talking about the transpiler stage, need to generate correct `module <name>` statements for all files.

I think you're looking for this:

https://github.com/adsharma/py2many/blob/main/py2many/toposort_modules.py#L19

Specifically, the top level AST node has a special attribute called __file__ which contains a path from which module name can be derived.

Our extensions to `ast.AST` are documented here: 

https://github.com/adsharma/py2many/blob/main/py2many/astx.py

But we haven't fixed up all the types yet, so type checkers still complain.

adsharma:
Also I just merged #423. Please rebase on top of it for the next CI run

tests/test_unsupported.py
```diff
@@ -113,15 +113,19 @@ def main(): r = Rectangle()
...
+    "bool_plus_int.v",
     "bool_to_int.jl",
     "bool_to_int.nim",
     "bool_to_int.rs",
+    "bool_to_int.v",
     "complex_dict.jl",
     "complex_dict.kt",
     "complex_dict.nim",
     "complex_dict.rs",
+    "complex_dict.v",  # ???
```
jayvdb, line 128: just a reminder that there are three(3) items in this file that need to be removed/revised/etc before this can be merged

div72: @jayvdb I assume you mean `complex_dict.v`, `dict_keys_explicit.v` and `dict_values.v`?

`complex_dict.v` was a false positive due to invalid transpilation of tuples, fixed now. The others are waiting vlang/v#10791 to be merged.

div72:
@adsharma `__file__` seems to be only contain the filename. I am simply appending `module main` at the top for every script right now, while that works for simple single-dir cases; it'll fail if there are sub-modules.

File imports and main function in a non-main module are both forbidden in V. Files in the submodule should also have the same name in the module statement as the directory that holds them. The problematic part is even if the path of the script is known, transpiler should be able to tell if it's working on a directory and if the any files on the directory has a main function to properly assign a module name.

adsharma:
Thank you @div72  and @nsauzede for your contributions!

tests/test_unsupported.py
spytheman: I just merged it.

PR #369, opened 2021-07-05
Two plugin fixes

jayvdb:


py2many/clike.py
```diff
@@ -561,6 +561,13 @@ def visit_IfExp(self, node):
 
     def _func_for_lookup(self, fname) -> Union[str, object]:
         func = class_for_typename(fname, None, self._imported_names)
+        if func is None:
+            return None
+        try:
+            hash(func)
+        except TypeError:
+            # Ignore unhashable, probably instance
```
adsharma, line 569: `logger.debug()` may be appropriate here so someone chasing this in the future may find a clue.

What test case is this breaking?

PR #371, opened 2021-07-05
Fix a few type check failures

adsharma:


adsharma:
@jayvdb I didn't notice your PRs before I created this one. But it looks like we tackled a different set of type errors. Let me know if there are any conflicting changes and I'll revert them.

py2many/exceptions.py
```diff
@@ -1,7 +1,7 @@
 import ast
 
 
-class AstErrorBase:
+class AstErrorBase(Exception):
```
jayvdb, line 4: this was intentionally not done so that subclasses need to explicitly choose which exception type they derive from, otherwise they are unthrowable.

py2many/ast_helpers.py
```diff
@@ -23,7 +23,7 @@ def create_ast_node(code, at_node=None):
 
 def create_ast_block(body, at_node=None):
```
jayvdb, line 24: this should be declared as `-> ast.If` (or `ASTxRewrittenIf`)?  Then its uses do not need a `cast`

py2many/rewriters.py
```diff
@@ -172,13 +172,17 @@ def visit_If(self, node):
                 rename(node.scopes[-2], "main", "main_func")
             # ast.parse produces a Module object that needs to be destructured
             if self.main_signature_arg_names == {"argc", "argv"}:
-                ret = ast.parse(
-                    "def main(argc: int, argv: List[str]) -> int: True"
-                ).body[0]
+                ret = cast(
+                    ast.FunctionDef,
+                    ast.parse("def main(argc: int, argv: List[str]) -> int: True").body[
```
jayvdb, line 177: IMO, anywhere this ugly `[ \n   0 \n    ]` appears, we need a helper.

py2many/rewriters.py
```diff
@@ -192,13 +196,15 @@ def __init__(self, language):
         super().__init__()
 
     def visit_JoinedStr(self, node):
-        new_node = ast.parse('"".join([])').body[0].value
+        new_node = cast(ast.Expr, ast.parse('"".join([])').body[0]).value
```
jayvdb, line 199: any `ast.parse(...).body[0]` should use `create_ast_node` whenever we're touching those pieces of code, and it looks like this PR is going to touch most of them.

jayvdb:
I don't see any major conflicting changes here.  If you want any of my PRs in before https://github.com/adsharma/py2many/pull/363 , let me know and i'll rebase and push them onto `main`

py2many/exceptions.py
adsharma: What does super() mean on line 8 without a base class?

py2many/rewriters.py
adsharma: Yeah - black makes it worse.

py2many/astx.py
```diff
@@ -0,0 +1,45 @@
+import ast
+from py2many.clike import LifeTime
```
jayvdb, line 2: create a new module for this and any other similar enums?

or move Lifetime here?

py2many/exceptions.py
jayvdb: All classes have `object` as their base.  In Python 3 that became implicit, but in Python 2 if there was no explicit `object` the class was a `type` instead of a `class`.
https://stackoverflow.com/questions/35958961/class-vs-type-in-python

py2many/rewriters.py
jayvdb: how has this been resolved?

py2many/rewriters.py
```diff
@@ -276,7 +283,8 @@ def __init__(self, language):
         self._language = language
 
     def _do_other_rewrite(self, node) -> ast.AST:
-        ifexpr = ast.parse("'True' if true else 'False'").body[0].value
+        ifexpr = cast(ast.Expr, ast.parse("'True' if true else 'False'").body[0]).value
```
jayvdb, line 286: Another `.body[0]`

py2many/inference.py
```diff
@@ -183,7 +184,7 @@ def visit_Constant(self, node):
     @staticmethod
     def _annotate(node, typename: str):
         # ast.parse produces a Module object that needs to be destructured
-        type_annotation = ast.parse(typename).body[0].value
+        type_annotation = cast(ast.Expr, ast.parse(typename).body[0]).value
```
jayvdb, line 187: Another `.body[0]`

py2many/exceptions.py
adsharma: >         super().__init__(msg)

The constructor of object doesn't take `msg` though, which is what the type checker was complaining about. 

We either remove this line, or have a base class that takes a message.

py2many/rewriters.py
adsharma: Missed it. Resolved in 68695c6

py2many/exceptions.py
jayvdb: ah, now I get you.  Well this code is correct, and and extra base doesnt make sense just to force subclasses to choice a more specific Exception subclass.  There are other ways to force that.

py2many/inference.py
jayvdb: still missed this one.

py2many/exceptions.py
adsharma: Plugins have the same issue. They assume that `self` is subclassed from `$LANGTranspiler`, but type checker complains because it doesn't see that.

Swift has the concept of `Extensions` to describe this pattern. We have to look into what python type checkers are willing to do to support something similar.

PR #375, opened 2021-07-05
Add two unsupported test cases

jayvdb:


py2many/inference.py
```diff
@@ -404,6 +404,16 @@ def visit_BinOp(self, node):
                 node.annotation = ast.Name(id="float")
             return node
 
+        # Container multiplication
```
adsharma, line 407: some of this is covered by `LEGAL_COMBINATIONS` further below. Perhaps moving the two pieces of code closer together could help the flow.

Also, `1 + 2j` errors out here. Perhaps make it legal while you're here?

jayvdb: Moved as requested.  And created https://github.com/adsharma/py2many/issues/379 for the unrelated `complex` set of binops, as tests also need creating for them.

adsharma: I see two instances of this code in the diff. Was it intentional?

py2many/inference.py
```diff
@@ -435,26 +445,32 @@ def visit_BinOp(self, node):
                     return node
             node.annotation = left
             return node
-        else:
```
adsharma, line 438: This code was covering `i8-i64 binop float` case. Perhaps didn't have test coverage.

py2many/inference.py
jayvdb: No. Resulted from multiple rebases.

py2many/inference.py
jayvdb: I believe no code path was changed here.  I only removed the unnecessary dangling else.

PR #383, opened 2021-07-06
CLI: comment unsupported

jayvdb:


py2many/clike.py
```diff
@@ -528,6 +528,42 @@ def visit_Delete(self, node):
...
+
+    def visit_AsyncFor(self, node):
+        target = self.visit(node.target)
+        iter = self.visit(node.iter)
+        return self.visit_unsupported_body(
+            node, f"async for {target} in {iter}", node.body
+        )
+
+    def visit_AsyncWith(self, node):
+        items = self.visit(node.items)
```
jayvdb, line 542: this should be `items = [self.visit(i) for i in node.items]`

Issue #379, opened 2021-07-05
Data type complex operations not supported


jayvdb:
2575b415 added `complex`, however all binops on it are rejected.

float, complex and int can be used together, with binops at least Add, Sub, Mult and Div.  Use of `complex` on either side of the binop results in a complex.

PR #387, opened 2021-07-06
python: allow complex binop int/float

adsharma:
Fixes: https://github.com/adsharma/py2many/issues/379

### DIFF
--- a/tests/cases/complex.py
+++ b/tests/cases/complex.py
@@ -1,0 +1,19 @@
+#!/usr/bin/env python3
+
+
+def complex_test():
+    c1 = 2 + 3j
+    c2 = 4 + 5j
+    c3 = c1 + c2
+    assert c3 == 6 + 8j
+    c4 = c1 + 3
+    assert c4 == 5 + 3j
+    c5 = c1 + 4j
+    assert c5 == 2 + 7j
+    c6 = c3 - 2.3
+    assert c6 == 3.7 + 8j
+
+
+if __name__ == "__main__":
+    complex_test()
+    print("OK")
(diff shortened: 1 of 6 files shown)


tests/cases/complex.py
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env python3
```
jayvdb, line 1: I'd prefer we dont have test cases here when none of the transpiler languages support the test case.
That means there are more failures when running in `UPDATE_EXPECTED=1` mode.

These could be just snippets in `test_unsupported` where they will fail, when they start working, requiring the lang be added to `EXPECTED_SUCCESSES`.

If you dont want to adjust this, I'll figure out some way to have test cases which are known to not work on any target lang, and probably also single out test cases which are also only working on Rust, as most of the Rust only test cases are very unlikely to start working on other targets without a lot of effort by some mythical new contributor.

adsharma: I got julia working.

adsharma:
The failure is caused by the interaction of py3.9 and astor.

```
x = ast.parse("3+4j")
astor.to_source(x)
'3 + 4.0j\n'
```

The test expects `3 + 4j`

adsharma:
Windows + py39 should have used `ast.unparse`. But the test failure suggests that it's using astor. I couldn't figure out from the logs why we end up with astor. Any clues?

jayvdb:
you could add this oddity to `standardise_python(..)`, replacing `.0j` with `j`.  Its not pretty, but would work for a subset of complex numbers, sufficient to test.

jayvdb (approved):

Issue #347, opened 2021-06-27
Add SonarCloud CI


jayvdb:
Mentioned at https://github.com/adsharma/py2many/issues/191#issuecomment-866498863 , SonarQube provides quite good analysis for a lot of languages.  There is a cloud CI provider, sonarcloud.io, which only provides the default/official plugins, which sadly excludes Rust, but it is still useful.  It also provides a way to record coverage; while it isnt as mature of codecov.io, it is quite good.

As can be seen at https://sonarcloud.io/summary/new_code?id=jayvdb_py2many it would have detected the problem being fixed by https://github.com/adsharma/py2many/pull/344

The automatic analysis is enabled at Administration -> Analysis Method, ignore the fact it is not recommended. Then I suggest setting the "New Code" to Number of days : `10`

We can disable code smells for test cases under Administration -> General Settings -> Analysis Scope setting "Source File Exclusions"

jayvdb:
I haven't found a way to change the default "Cognitive Complexity" from 15 to say 20.  From my use of SonarCloud at work, this shouldnt cause a problem, as new code in PRs wont fail unless they are modifying the functions that are currently over the Cognitive Complexity threshold.

jayvdb:
https://sonarcloud.io/summary/overall?id=jayvdb_py2many now has a C rating for reliability, and A for maintainability (165 smells, with only a few in `tests/test_*.py` so I haven't excluded those).

To achieve that, I set the exclusions to
- `tests/cases/**`
- `tests/ext_cases/**`
- `tests/expected/**`
- `tests/ext_expected/**`

jayvdb:
https://sonarcloud.io/project/overview?id=jayvdb_py2many updated, with `tests/dir_cases/**` added to exclusions

adsharma:
The "C" reliability rating seems to be coming from two workarounds to silence pyflakes.

Which brings up the question of why not use flake8, which runs pyflakes internally and offers more flexible error checking (such as #noqa comments)?

Issue #149, opened 2021-04-30
v0.2.0 release


adsharma:
Most other languages are passing tests now:

SKIPPED [1] tests/test_cli.py:64: infer_ops.kt doesnt compile
SKIPPED [1] tests/test_cli.py:64: infer_ops.nim doesnt compile
SKIPPED [1] tests/test_cli.py:64: expected/str_enum.cpp not found
SKIPPED [1] tests/test_cli.py:64: expected/str_enum.go not found
============ 189 passed, 4 skipped, 5 warnings in 388.63s (0:06:28) ============

79% code coverage

C++/go need string enum support (type inference is done)
Kotlin/Nim needs some type casts (copy-pasta from rust should fix).

adsharma:
@jayvdb first test run with no skips and no failures since you set up CI

https://github.com/adsharma/py2many/pull/156/checks?check_run_id=2494093580

================= 193 passed, 5 warnings in 309.69s (0:05:09) ==================

jayvdb:
Well, that is easy to fix - I have lots of test cases that were not passing for various languages.

And incorporate grumpy test suite https://github.com/adsharma/py2many/issues/116 ;-)

adsharma:
I'm thinking of tagging the tree with 0.2 and uploading a new version to pypi

jayvdb:
Up to you whether you want to merge https://github.com/adsharma/py2many/pull/214 into v0.2.0 .  It fixed the code to work with g++, which seems worth including.

I'd appreciate it if you could also pull in https://github.com/john-michaelburke/py2many/commit/8b7574ad46579405ebed8af386028518d9a692db before cutting the release.  Then my internal project using py2many can use the released version.

adsharma:
https://pypi.org/project/py2many/0.2.1/

PR #388, opened 2021-07-07
Changelog

adsharma:


CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+## Unreleased
+
+### Added
+- Migrate API translation code to plugin infra
+- Support for self transpilation with improved error reporting
```
jayvdb, line 5: this is just a testing thing; has no outside benefit as self-transpilation renders junk.

CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
...
+- New target: python. Transpiles untyped python to typed python.
+- Directory mode. Cross module type inference.
```
jayvdb, line 7: "Directory mode" previously existed.

Is there a test which does "Cross module type inference." ?  If they exist, there are not many of them, so I'd be wanting to create a few more tests to verify it works reasonably well.

CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
...
+- Experimental: pyO3 extension for rust
```
jayvdb, line 8: Perhaps pyO3 mode could be converted into a new target "pyO3" or "rust+pyO3" ?

CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
...
+- Rust: python sealed classes -> Rust Enum
```
jayvdb, line 9: i think we need to hyperlink sealed to adt repo.

CHANGELOG.md
adsharma: Opened #391

CHANGELOG.md
adsharma: Will remove

CHANGELOG.md
adsharma: It's a sub-target of rust, not a standalone one. A large module with a small interface to python would be mostly the same as rust. Only the code that interfaces with python would be transpiled differently.

jayvdb:
https://github.com/adsharma/py2many/issues/149#issue-873158088 said we had "79% code coverage", but https://github.com/adsharma/py2many/runs/2571696956#step:12:518 says 84%

last `main` commit says 90% https://github.com/adsharma/py2many/runs/3005235149#step:16:748 `\*/` . That is the magic threshold for me.  And reaching 90% means we dont even need to say what the prior coverage level was, as anyone who cares about this line in the changelog will know each percentage point in that region is hard work.

And we should now start requiring 90%.  Note that if we integrate SonarCloud (https://github.com/adsharma/py2many/issues/347), we can make that a soft requirement defined in the "Quality Gate", and it will fail the PR, but that can be ignored when merging to `main` (and ideally then followed by another PR which boosts the coverage back over the threshold).

jayvdb:
new test cases are IMO all new working functionality worth putting in changelog.  fib_with_argparse/sealed/with_open are all Rust-only, and IMO Rust should have a separate sub-section in the relnotes given it usually gets features first.

```
❯ git diff --name-only --diff-filter=A v0.2.1..HEAD -- tests/cases
      1 tests/cases/bitops.py
      2 tests/cases/built_ins.py
      3 tests/cases/byte_literals.py
      4 tests/cases/comment_unsupported.py
      5 tests/cases/dict.py
      6 tests/cases/fib_with_argparse.py
      7 tests/cases/sealed.py
      8 tests/cases/sys_argv.py
      9 tests/cases/sys_exit.py
     10 tests/cases/with_open.py
```

Test successes since the last release
```
❯ GIT_PAGER=cat git diff --name-only --diff-filter=A v0.2.1..HEAD -- tests/expected | \grep --color=always -v '\.py' | cut -d '.' -f 1 | uniq -c | sort -rn
   7 tests/expected/sys_exit  # all langs
   7 tests/expected/sys_argv
   7 tests/expected/comment_unsupported
   7 tests/expected/built_ins  # min() was fixed
   7 tests/expected/bitops
   4 tests/expected/nested_dict  # rust & julia & go & kotlin
   3 tests/expected/dict # rust & julia & dart
   2 tests/expected/byte_literals # rust & julia
   1 tests/expected/with_open  # rust
   1 tests/expected/sealed  # rust
   1 tests/expected/fib_with_argparse  # rust
```

PR #389, opened 2021-07-07
visit_ImportFrom fix and NamedExpr not implemented

jayvdb:
AttributeError occurs because node.module is None
when `from . import ...`

### DIFF
--- a/py2many/clike.py
+++ b/py2many/clike.py
@@ -527,6 +527,10 @@ class CLikeTranspiler(ast.NodeVisitor):
                 f"{name} unimplemented on line {node.lineno}:{node.col_offset}"
             )
 
+    def visit_NamedExpr(self, node):
+        target = self.visit(node.target)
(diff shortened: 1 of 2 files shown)


py2many/clike.py
```diff
@@ -527,6 +527,10 @@ def visit_unsupported_body(self, node, name, body):
                 f"{name} unimplemented on line {node.lineno}:{node.col_offset}"
             )
 
+    def visit_NamedExpr(self, node):
+        target = self.visit(node.target)
```
adsharma, line 531: visit node.value as well?

adsharma (commented):
https://www.python.org/dev/peps/pep-0572/

adsharma (commented):
typo on Pythoon

py2many/clike.py
jayvdb: I've passed that as the `body` parameter to `visit_unsupported_body`, similar to the other uses.
I expect we'll enhance `visit_unsupported_body` to do that visiting.

Issue #391, opened 2021-07-07
Add a test case for directory mode with cross module type inference


adsharma:
A minimal test case + a realistic test case would be ideal

adsharma:
Minimal test case is all we have for now. Hopefully we will find end users who run this on non-trivial code and file bugs.

PR #390, opened 2021-07-07
_dispatch: Ignore IndexError in plugins

jayvdb:
clike: Ignore use of types without brackets

Issue #339, opened 2021-06-26
Windows CI doesnt invoke Julia and Kotlin formatters


jayvdb:
Recently discussed at https://github.com/adsharma/py2many/pull/336#discussion_r659010567

This is why `test_self_transpile.py` allows `FileNotFoundError` for Kotlin and Julia.

div72:
> And I suspect that the newly added V and smt also dont support Windows

Is there any Windows specific checks for py2many? V should work fine otherwise.

jayvdb:
https://github.com/domluna/JuliaFormatter.jl/pull/599 and https://github.com/pinterest/ktlint/issues/1363 look like good solutions which would avoid needing to build weird hacks for CI.

jayvdb:
SMT and Kotlin have been addressed by using jgo.

More broadly SMT is covered by https://github.com/py2many/py2many/issues/568

I believe the Julia issues have also been resolved by https://github.com/py2many/py2many/pull/479

There are no Windows specific code in py2many, but there are definitely some specific Vlang specified problems being avoided by `_suppress_exceptions=(AstNotImplementedError, AstUnrecognisedBinOp),` at https://github.com/py2many/py2many/blob/c12df6a/tests/test_transpile_self.py#L254 , but a specified issue about that is worth creating and tackling but is not this issue.

https://github.com/py2many/py2many/pull/569 removes the workarounds which were put in place that creates this issue, so this can be closed if that succeeds.

PR #396, opened 2021-07-08
tests: add a test for directory mode

adsharma:
Related to: https://github.com/adsharma/py2many/issues/391

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -119,7 +119,7 @@ jobs:
         run: pip install black pyflakes cpplint
 
       - name: Remove tests/cases/*.py & tests/expected/*.py
-        run: rm tests/cases/*.py tests/expected/*.py
+        run: rm tests/cases/*.py tests/expected/*.py tests/dir_cases/test1-*-expected/*.py
 
       - name: Run black
         run: black --check */ *.py
--- a/pygo/transpiler.py
+++ b/pygo/transpiler.py
@@ -491,9 +491,9 @@ class GoTranspiler(CLikeTranspiler):
             if lookup in MODULE_DISPATCH_TABLE:
                 go_module_name, go_name = MODULE_DISPATCH_TABLE[lookup]
                 go_module_name = go_module_name.replace(".", "/")
-                return f'import ("{go_module_name}/{go_name}"")'
+                return f'import ("{go_module_name}/{go_name}")'
         module_name = module_name.replace(".", "/")
-        return "\n".join([f'import ("{module_name}/{name}"")' for name in names])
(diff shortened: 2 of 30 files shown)


adsharma:
Not sure how tests/expected/*.py are ignored in pyflakes. I should probably use the same mechanism.

jayvdb:
`rm tests/cases/*.py tests/expected/*.py` in https://github.com/adsharma/py2many/blob/main/.github/workflows/main.yml#L121-L122

adsharma:
go, julia, kotlin, nim and rust fail the newly added test, only on windows.
Failure mode: no files in the target dir.

I first thought if it's an issue with handling paths. In that case - cpp and dart should fail too. But they pass.

jayvdb:
IMO, raise `SkipTest` on Windows for this test on any that dont pass.

We already have bad Windows support, and this is just another issue to add to it.

Lets get this in and fix release blockers.

jayvdb:
for reference, https://github.com/adsharma/py2many/issues/339 is the issue about Julia and Kotlin failing on Windows.

btw, https://github.com/adsharma/py2many/pull/390 is ready

adsharma:
Noting the debug information here and going with your suggestion of SkipTest:

```
>           assert output.diff_files == []
E           AssertionError: assert ['bar.go', 'baz.go', 'foo.go'] == []
E             Left contains 3 more items, first extra item: 'bar.go'
E             Full diff:
E             - []
E             + ['bar.go', 'baz.go', 'foo.go']

D:\a\py2many\py2many\tests\test_cli.py:447: AssertionError
---------------------------- Captured stdout call -----------------------------
Transpiling whole directory to D:\a\py2many\py2many\tests\dir_cases\test1-go-generated:
writing to: D:\a\py2many\py2many\tests\dir_cases\test1-go-generated\bar.go 47 bytes
writing to: D:\a\py2many\py2many\tests\dir_cases\test1-go-generated\bar.go done
writing to: D:\a\py2many\py2many\tests\dir_cases\test1-go-generated\baz.go 54 bytes
writing to: D:\a\py2many\py2many\tests\dir_cases\test1-go-generated\baz.go done
writing to: D:\a\py2many\py2many\tests\dir_cases\test1-go-generated\foo.go 168 bytes
writing to: D:\a\py2many\py2many\tests\dir_cases\test1-go-generated\foo.go done

Finished!
Successful: 3
Failed to convert: 0
```

adsharma:
It certainly looks like a race condition, where py2many/cli.py:_format_one() is still running the formatter in the background, even after `run()` has returned. Documentation says it returns only after the process has completed.

Working around by ignoring exceptions inside `_recursive_rmdir`

Issue #402, opened 2021-07-08
go: Could not infer: None


jayvdb:
The following results in exception `Could not infer: None` only on Go.
```py
class Foo:
    def bar(self):
        return self.x()
```
This is an intentional error, to force the lack of inference for `self.x`.

All languages other than Go transpile this, and then the compiler/invoker fails.  Go fails to transpile this.  IMO Go should be consistent.

PR #403, opened 2021-07-08
pygo: Provide a default return if not inferred 

jayvdb:
Fixes https://github.com/adsharma/py2many/issues/402

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,6 +54,11 @@ test = true
 path = "tests/expected/built_ins.rs"
 
 [[bin]]
+name = "cls"
+test = true
+path = "tests/expected/cls.rs"
+
+[[bin]]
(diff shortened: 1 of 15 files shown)


jayvdb:
This still fails `infer_ops`.  I'll get to that soon.  But I dont mind if this doesnt make it into v0.3.0 ; it is only replacing one problem with a different more predicable problem.

jayvdb:
After this change, the only gofmt error for the sample on https://github.com/adsharma/py2many/issues/402 is
```
./cls.go:11:9: undefined: x
```

And now most of the go self transpilation is successful except for gofmt errors.

adsharma:
Looks good. Perhaps rebase on top of #396 ?

adsharma:
Closed via 0a043bd

PR #406, opened 2021-07-09
Allow import typing

jayvdb:
Previously only `from typing import ..` was supported.

### DIFF
--- a/py2many/clike.py
+++ b/py2many/clike.py
@@ -210,6 +210,8 @@ class CLikeTranspiler(ast.NodeVisitor):
             return [self._typename_from_type_node(e) for e in node.elts]
         elif isinstance(node, ast.Attribute):
             node_id = get_id(node)
+            if node_id.startswith("typing."):
+                node_id = node_id[7:]
             return node_id
         elif isinstance(node, ast.Subscript):
             # Store a tuple like (List, int) or (Dict, (str, int)) for container types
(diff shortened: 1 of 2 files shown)


py2many/clike.py
```diff
@@ -208,6 +208,11 @@ def _typename_from_type_node(self, node) -> Union[List, str, None]:
             return get_id(node)
         elif isinstance(node, ast.Tuple):
             return [self._typename_from_type_node(e) for e in node.elts]
+        elif isinstance(node, ast.Attribute):
+            node_id = get_id(node)
+            if node_id.startswith("typing."):
```
adsharma, line 213: One could also write:

```
from typing import List as typ_list
```
A more general solution would involve some type of eval/__import__ of the statement above and then looking up typing.List in a dict, like what we're doing in class_for_typename/_func_for_lookup.

Could use remove_prefix post 3.9:

https://www.python.org/dev/peps/pep-0616/

tests/cases/binit.py
```diff
@@ -2,10 +2,10 @@
 
 # http://rosettacode.org/wiki/Bin_given_limits
 
-from typing import List
```
adsharma, line 5: For the case where only one thing is imported, the old code was more concise.
Why not add a new test and leave these cases alone?

jayvdb: It isnt more consise
```
from typing import List;List
import typing;typing.List
```
Not sure why you dont want to use the existing test cases.  A new cases just for this import is hardly worth it, and ive run out of time I can allocate to this project this weekend.

py2many/clike.py
jayvdb: sure.  there are lots of problems, and this only solves one of them, and does so without the overhead of eval/import.  Created https://github.com/adsharma/py2many/issues/409 for a fallback to more difficult cases.

PR #416, opened 2021-07-12
pyrs: Fix while True

jayvdb:
Rust complains about `while true`, requires `loop { .. }`.


jayvdb:
The macos build failed Running tox and has no logs. :/
@adsharma, Would be great if we could restart just that failing job.  I dont have permission to do that.

adsharma:
Thanks for adding the tests!

Issue #417, opened 2021-07-12
AttributeError: 'FunctionDef' object has no attribute 'assigned_from'


jayvdb:
The following causes a `AttributeError: 'FunctionDef' object has no attribute 'assigned_from'` at `bar.append(1)`
```py
class Foo:
    def bar(self):
        pass

    def baz(self):
        bar = []
        bar.append(1)
```
It is getting confused between the local var and class method.

After adding some detection:

```py
/Users/john.vandenberg/python/py2many/py2many/cli.py:150: in _transpile_one
    tree, infer_meta = core_transformers(tree, trees)
/Users/john.vandenberg/python/py2many/py2many/cli.py:74: in core_transformers
    add_list_calls(tree)
/Users/john.vandenberg/python/py2many/py2many/context.py:8: in add_list_calls
    return ListCallTransformer().visit(node)
/usr/local/Cellar/python@3.9/3.9.5/Frameworks/Python.framework/Versions/3.9/lib/python3.9/ast.py:407: in visit
    return visitor(node)
/usr/local/Cellar/python@3.9/3.9.5/Frameworks/Python.framework/Versions/3.9/lib/python3.9/ast.py:483: in generic_visit
    value = self.visit(value)
/usr/local/Cellar/python@3.9/3.9.5/Frameworks/Python.framework/Versions/3.9/lib/python3.9/ast.py:407: in visit
    return visitor(node)
/usr/local/Cellar/python@3.9/3.9.5/Frameworks/Python.framework/Versions/3.9/lib/python3.9/ast.py:483: in generic_visit
    value = self.visit(value)
/usr/local/Cellar/python@3.9/3.9.5/Frameworks/Python.framework/Versions/3.9/lib/python3.9/ast.py:407: in visit
    return visitor(node)
/usr/local/Cellar/python@3.9/3.9.5/Frameworks/Python.framework/Versions/3.9/lib/python3.9/ast.py:483: in generic_visit
    value = self.visit(value)
/usr/local/Cellar/python@3.9/3.9.5/Frameworks/Python.framework/Versions/3.9/lib/python3.9/ast.py:407: in visit
    return visitor(node)
/usr/local/Cellar/python@3.9/3.9.5/Frameworks/Python.framework/Versions/3.9/lib/python3.9/ast.py:492: in generic_visit
    new_node = self.visit(old_value)
[...]

PR #418, opened 2021-07-12
analysis: fix the order of lookup

adsharma:
We need to search the scopes in reverse order. It exposed a second
bug. Which is that a variable defined in the outer scope is redefined
in the current scope according to the current logic in visit_Assign_one.

Fix the logic to define it only if it hasn't been defined before.

Fixes: https://github.com/adsharma/py2many/issues/417

### DIFF
--- a/py2many/scope.py
+++ b/py2many/scope.py
@@ -52,7 +52,7 @@ class ScopeList(list):
                 if get_id(var) == lookup:
                     return var
 
-        for scope in self:
+        for scope in reversed(self):
             defn = None
             if not defn and hasattr(scope, "vars"):
                 defn = find_definition(scope, "vars")
@@ -71,6 +71,10 @@ class ScopeList(list):
                 for imp in scope.imports:
                     if imp.name == lookup:
                         return imp
+
+    @property
+    def parent_scopes(self):
+        return ScopeList(self[:-1])
 
 
 class ScopeTransformer(ast.NodeTransformer, ScopeMixin):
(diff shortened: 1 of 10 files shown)


jayvdb:
Those two failures ring a bell.  It looks like https://github.com/adsharma/py2many/pull/363/files#diff-7dce7c041b060d90c8f5d5d04c7114f603028662343340b6a59eb8d00d7bf95cR28 , or something else in that PR, is causing these two failures.

And I would like a sample / test for the extra bit being fixed here, so I can understand it better.  The new code in the `_visit_AssignOne`'s isnt immediately looking obvious to me in what it is achieving, but I'll start playing with it in the hope that it makes more sense.

> Which is that a variable defined in the outer scope is redefined in the current scope according to the current logic in visit_Assign_one.

py2many/scope.py
```diff
@@ -72,6 +72,9 @@ def find_import(self, lookup):
                     if imp.name == lookup:
                         return imp
 
+    def parent(self):
```
jayvdb, line 75: IMO this should be `parent_scopes` as `parent` sounds like it returns a single scope, and could be a `@property`.

adsharma:
If you undo the changes to transpiler.py, you'll see failures on binit.py,
which explains what the bug is.

adsharma:
Also this logic could be called `defined_before` which is already there in the code base. But it's based on line numbers, which is not as robust as scope based logic.

Issue #410, opened 2021-07-10
Redundant types results in differences in output


jayvdb:
The following two should be identical as an easy case of type inference
```py
a9: List[str] = ["a", "b", "c", "d"]
```
and
```py
a9 = ["a", "b", "c", "d"]
```
Both pass in all langs, however the output is quite different for each.

```diff
diff --git a/tests/cases/coverage.py b/tests/cases/coverage.py
index 936d618..fcc4828 100755
--- a/tests/cases/coverage.py
+++ b/tests/cases/coverage.py
@@ -55,7 +55,7 @@ def show():
     # lists, sets and dict
     a5 = [1, 2, 3]
     print(len(a5))
-    a9: List[str] = ["a", "b", "c", "d"]
+    a9 = ["a", "b", "c", "d"]
     print(len(a9))
     a6 = {1, 2, 3, 4}
     print(len(a6))
diff --git a/tests/expected/coverage.cpp b/tests/expected/coverage.cpp
index 7120f11..c46bfc2 100644
--- a/tests/expected/coverage.cpp
+++ b/tests/expected/coverage.cpp
@@ -64,8 +64,8 @@ inline void show() {
   std::vector<int> a5 = {1, 2, 3};
   std::cout << a5.size();
   std::cout << std::endl;
-  std::vector<std::string> a9 = std::vector<std::string>{
-      std::string{"a"}, std::string{"b"}, std::string{"c"}, std::string{"d"}};
+  std::vector<std::string> a9 = {std::string{"a"}, std::string{"b"},
+                                 std::string{"c"}, std::string{"d"}};
   std::cout << a9.size();
   std::cout << std::endl;
   std::set<int> a6 = std::set<int>{1, 2, 3, 4};
diff --git a/tests/expected/coverage.dart b/tests/expected/coverage.dart
index 92802a4..2a27fc9 100644
--- a/tests/expected/coverage.dart
+++ b/tests/expected/coverage.dart
@@ -49,7 +49,7 @@ show() {
   print(sprintf("%s", [sum1]));
   final a5 = [1, 2, 3];
   print(sprintf("%s", [a5.length]));
-  List<String> a9 = ["a", "b", "c", "d"];
+  final a9 = ["a", "b", "c", "d"];
   print(sprintf("%s", [a9.length]));
   final Set<int> a6 = new Set.from([1, 2, 3, 4]);
   print(sprintf("%s", [a6.length]));
[...]

PR #420, opened 2021-07-13
Simplify assign one

adsharma:
Related to https://github.com/adsharma/py2many/issues/410

### DIFF
--- a/pycpp/transpiler.py
+++ b/pycpp/transpiler.py
@@ -510,7 +510,7 @@ class CppTranspiler(CLikeTranspiler):
             if "auto" in typename:
                 return f"{{{elements_str}}}"
             return f"{typename}{{{elements_str}}}"
-        return f"std::vector<{element_type}>{{{elements_str}}}"
+        return f"{{{elements_str}}}"
 
     def visit_Set(self, node):
         self._usings.add("<set>")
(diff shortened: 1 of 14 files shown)


jayvdb (commented):
https://github.com/adsharma/py2many/commit/6378f1598b3f8cc3822f95464ba18ee04ed2f9d4 references #410, but it seems to be unrelated, as it is seemingly/just re-blaming code with any improvement to the output.

The issue is that coverage.py should be
```diff
-a9: List[str] = ["a", "b", "c", "d"]
+a9 = ["a", "b", "c", "d"]
```

And then the problem is fixed if `coverage.rs` does not include this:
```diff
-    let a9: Vec<&str> = vec!["a", "b", "c", "d"];
+    let a9: &Vec<&str> = &vec!["a", "b", "c", "d"];
```

adsharma:
It's deleting the special code paths for lists, since the normal code paths is general enough.
This diff stack trims around ~40 lines.

Re: the problem python, julia, kotlin have

They add a type annotation only in `visit_AnnAssign`, even if there is an inferred one.
The fix should be to change `visit_AssignOne` to add the annotation just like pyrs does.

That leaves a few minor issues:

* final vs no final on dart (I think final is more correct)
* Reference vs no reference on Rust

adsharma:
Correction: only julia and kotlin have this problem.

Python is ok and handled by InferredAnnAssignRewriter

adsharma:
One option is to add this rewriter to other languages that need it:

https://paste.ubuntu.com/p/W3Sm6k5cZQ/

But right now it opens up other problems that I'm not sure we can solve before v0.3

jayvdb (approved):

Issue #356, opened 2021-06-29
v0.3


jayvdb:
Once https://github.com/adsharma/py2many/pull/350 is merged, I think we're due for a 0.3 release.

Any features wanted before a new release?

ping @PXshadow @dstromberg @john-michaelburke @dumblob 

@nsauzede , I notice you have an active fork with some changes.  Any you feel are close to being ready - we can help you get them rebased and merged if you let us know which branches we should be investigating.

john-michaelburke:
Nothing on my end.

adsharma:
#350 is merged. I'm taking a few days off. Will be back online this weekend.

nsauzede:
@jayvdb hey thank for reaching out
Yes, I did some quick experiment to introduce early V support (vlang)
but I just tried a quick rebase on latest master and indeed there are conflicts
I'll try to see if I can sort them out...
else yes it would be nice to have a bit of help
the branch of interest is `f-v`

jayvdb:
Looks like you are already past the conflicts.  I've added some further guidance at https://github.com/adsharma/py2many/issues/319#issuecomment-870425270 .  ping me again on that issue if you're getting stuck.

nsauzede:
> 
> 
> Looks like you are already past the conflicts. I've added some further guidance at [#319 (comment)](https://github.com/adsharma/py2many/issues/319#issuecomment-870425270) . ping me again on that issue if you're getting stuck.

Yes, but @div72 did a better job (including my initial work) here : https://github.com/adsharma/py2many/pull/357
Please consider reviewing/approving/merging his instead

PXshadow:
It's looking great from my view, congrats on this release, I'm exicted to see a transpiled out typed python source, I think it could be very useful. Still a decent amount to do on my end to transpile the required Go libraries for the Go target of py2many.

adsharma:
#355 is a good one to close. Otherwise it looks like we're in good shape!

jayvdb:
https://github.com/adsharma/py2many/blob/d5ef9c3/py2many/rewriters.py#L26-L27 was some debugging left behind by me, and should be removed.

I think we should start a changelog.

jayvdb:
In addition to https://github.com/adsharma/py2many/pull/363, I have another bunch of fixes on the way.

adsharma:
Looks like we'll have most of the outstanding fixes in today. Sent a changelog PR as well.

Should we look into vlang after the release or hold it till #357 is passing tests?

[...]

PR #421, opened 2021-07-13
cli: Restore _suppress_exceptions

jayvdb:
Also update the self-transpile tests to reflect
that cpp, julia and nim no longer have exceptions
occur due to recent bug fixes.
Only go still has exceptions during self-transpilation.


jayvdb:
@adsharma, ping, this is the last regression that I believe is blocking https://github.com/adsharma/py2many/issues/356

py2many/clike.py
```diff
@@ -164,10 +164,13 @@ def _cast(self, name: str, to) -> str:
...
         if sys.version_info < (3, 9, 0):
-            if not isinstance(node.slice, ast.Index):
-                raise AstNotImplementedError("Advanced Slicing not supported", node)
-            slice_value = node.slice.value
+            if isinstance(node.slice, ast.Index):
+                slice_value = node.slice.value
+            else:
+                slice_value = node.slice
         else:
+            if isinstance(node.slice, ast.Slice):
```
adsharma, line 172: Left a comment on https://github.com/adsharma/py2many/issues/422 on this topic.

py2many/cli.py
```diff
@@ -352,7 +352,10 @@ def _relative_to_cwd(absolute_path):
 
 
 def _get_output_path(filename, ext, outdir):
-    output_path = outdir / (filename.stem + ext)
+    directory = outdir.joinpath(str(filename.parent))
```
adsharma, line 355: `outdir / filename.parent`

py2many/cli.py
```diff
@@ -367,7 +370,7 @@ def _process_one(
     Returns False if reformatter failed.
     """
     suffix = f".{args.suffix}" if args.suffix is not None else settings.ext
-    output_path = _get_output_path(filename, suffix, outdir)
+    output_path = _get_output_path(pathlib.Path(filename.name), suffix, outdir)
```
adsharma, line 373: is this change necessary? filename is already a Path.

adsharma (commented):
The first three diffs in the stack can be merged with minor changes.
Seeking community input for the advanced slicing. Will report back if I hear anything.

py2many/cli.py
jayvdb: filename can be an absolute path, which breaks the algorithm in `_get_output_path`

py2many/clike.py
jayvdb: This change here is only to make v0.3.0 consistent across Python versions.  I expect this will be supported at some stage, but it isnt worth holding up v0.3.0 to implement it.

py2many/cli.py
```diff
@@ -81,7 +81,10 @@ def core_transformers(tree, trees):
 
 
 def _transpile(
-    filenames: List[pathlib.Path], sources: List[str], settings: LanguageSettings
+    filenames: List[pathlib.Path],
+    sources: List[str],
+    settings: LanguageSettings,
+    _suppress_exceptions=Exception,
```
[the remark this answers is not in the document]
jayvdb: This is a hidden arg, used only to support our self-transpile tests, and becomes unnecessary and can be removed once all languages can parse Python code without throwing exceptions.  We're getting quite close.  To me it is ridiculous to add typing here, especially as there are so many other typing problems to fix.

py2many/cli.py

jayvdb:
ping @adsharma

py2many/toposort_modules.py
```diff
@@ -7,7 +7,7 @@
 
 def module_for_path(path: Path) -> str:
     # strip out .py at the end
-    return ".".join(path.parts)[:-3]
```
adsharma, line 10: This doesn't do the same thing as before:

```
>>> module_for_path(Path("a/b/c.py"))
'a.b.c'
>>> module_for_path_after(Path("a/b/c.py"))
'a/b/c'
```

py2many/cli.py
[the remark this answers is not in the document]
adsharma: It wasn't so much about internal/external or temporary/long term change. The most surprising thing about it was it could be a bool or a list of exceptions. It could be resolved with a comment, but type annotations work better sometimes.

Other mitigations are possible - split it into two variables or interpret empty list as boolean False etc.

py2many/cli.py
adsharma: `filename.relative_to(filename.parent)` makes the logic clearer

adsharma (approved):
Minor nits. I'll merge and fix it up.

py2many/toposort_modules.py
jayvdb: curious that there were no tests failing.

adsharma:
It's used only for dependency management in toposort_modules.py, so both
variants work.
But it's possible that we start using module_for_path() for other stuff in
which case this becomes significant.

PR #423, opened 2021-07-13
context: add a transformer to annotate lhs

adsharma:
The left hand side (lhs) and right hand side (rhs) need different
treatment in the transpiler. Having this info handy would help 
with cases discussed in

https://github.com/adsharma/py2many/pull/294

### DIFF
--- a/py2many/astx.py
+++ b/py2many/astx.py
@@ -47,5 +47,6 @@ class ASTxIf(ast.If):
 class ASTx(ast.AST):
     annotation: ASTxName
     rewritten: bool = False
+    lhs: bool = False
     scopes: List["ASTx"] = field(default_factory=list)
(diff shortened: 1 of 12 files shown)


jayvdb (approved):

jayvdb (commented):
bah, this doesnt do what https://github.com/adsharma/py2many/pull/294 "needs", because it only cares about LHS vs RHS.

jayvdb (changes requested):
per above. this isnt helpful

adsharma:
There was a bug that's fixed by the updated commit. I tested that this code does the same thing as #294 and finds a couple of extra places where we needed to throw. Here's what I used to test:

https://paste.ubuntu.com/p/Bd726WWkKt/

PR #425, opened 2021-07-14
Cleanups

adsharma:
Nits from #421

### DIFF
--- a/py2many/cli.py
+++ b/py2many/cli.py
@@ -2,12 +2,12 @@ import argparse
 import ast
 import functools
 import os
-import pathlib
 import sys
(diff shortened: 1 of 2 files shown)


jayvdb:
Now you can see why `relative_to` isnt so great.

adsharma:
Looks like this is one place where os.path.relpath and relative_to differ:

https://stackoverflow.com/questions/38083555/using-pathlibs-relative-to-for-directories-on-the-same-level

I restored the `_relative_to_cwd()` code, which is actually a better fix than the ones discussed on stackoverflow.

Issue #426, opened 2021-07-14
Remove `raise Exception(..)`


jayvdb:
Many of these should be `AstNotImplementedError`.

`raise()` related ones could possibly be merged into a helper, and if it is absolutely only catching coding errors, a new exception type might be more appropriate, e.g. `AstInvalidBuiltinInvocation`.  But IMO we shouldnt be detecting bad code except in very limited scenarios (e.g. ctypes integer bounds we are doing, but mypy rejects those anyway per https://github.com/adsharma/py2many/issues/359 ).  If we're detecting bad code, we need to somehow create that bad code in order to test that part of the code, and such bad code  cant be run on Python to demo that the test case is valid, and that seems like lots of brittle busy-work.

```py
❯ GIT_PAGER=cat git grep -n --after-context=1 'raise.*Exception'
py2many/clike.py:610:        raise Exception("Unimplemented")
py2many/clike.py-611-
--
py2many/clike.py:613:        raise Exception("Unimplemented")
py2many/clike.py-614-
--
py2many/clike.py:616:        raise Exception("Unimplemented")
py2many/clike.py-617-
--
pydart/plugins.py:105:            raise Exception(
pydart/plugins.py-106-                "encountered range() call with unknown parameters: range({})".format(
--
pygo/plugins.py:103:        raise Exception(
pygo/plugins.py-104-            "encountered range() call with unknown parameters: range({})".format(vargs)
--
pyjl/plugins.py:99:        raise Exception(
pyjl/plugins.py-100-            "encountered range() call with unknown parameters: range({})".format(vargs)
--
pyjl/transpiler.py:177:        raise Exception(
pyjl/transpiler.py-178-            "encountered range() call with unknown parameters: range({})".format(vargs)
--
pykt/plugins.py:101:        raise Exception(
pykt/plugins.py-102-            "encountered range() call with unknown parameters: range({})".format(vargs)
--
pynim/plugins.py:47:        raise Exception(
pynim/plugins.py-48-            "encountered range() call with unknown parameters: range({})".format(vargs)
--
pyrs/plugins.py:84:        raise Exception("read() with more than one argument")
pyrs/plugins.py-85-
--
pyrs/plugins.py:93:        raise Exception("write() with more than one argument")
pyrs/plugins.py-94-
--
pyrs/plugins.py:120:        raise Exception(
pyrs/plugins.py-121-            "encountered range() call with unknown parameters: range({})".format(vargs)
```

adsharma:
[...]

PR #428, opened 2021-07-14
Remove `raise Exception`

jayvdb:
Closes https://github.com/adsharma/py2many/issues/426

### DIFF
--- a/pyjl/plugins.py
+++ b/pyjl/plugins.py
@@ -89,16 +89,12 @@ class JuiliaTranspilerPlugins:
         return cls
 
     def visit_range(self, node, vargs: List[str]) -> str:
+        assert 0 < len(node.args) < 4, "range() call with unexpected arguments"
         if len(node.args) == 1:
             return f"(0:{vargs[0]} - 1)"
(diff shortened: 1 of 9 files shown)


pyjl/plugins.py
```diff
@@ -89,16 +89,12 @@ def visit_ap_dataclass(self, cls):
         return cls
 
     def visit_range(self, node, vargs: List[str]) -> str:
+        assert 0 < len(node.args) < 4, "range() call with unexpected arguments"
```
jayvdb, line 92: Do you want me to remove the `assert` or not?

adsharma: My suggestion is to keep it, convert the assert to a `raise AppropriateException` and then implement a flag in `clike._dispatch()` to catch an entire class of exceptions and let the transpilation proceed.

This way: `range(1,2,3,4)` would get transpiled into the target language under the flag (fails later in the pipeline), but allows us to discover more problems in the source. The flag itself can be another PR if you prefer.

adsharma:
Needs rebase and further discussion

PR #430, opened 2021-07-14
Release v0.3

adsharma:


jayvdb (changes requested):
sdists should include tests.  bdists/wheels should not, and are what most users will consume.  This is a deal breaker for me.

adsharma:
Yes - that makes sense. I was mainly concerned about the wheels and other bdists.

jayvdb (approved):

PR #431, opened 2021-07-14
ci: Run CI on all branches

div72:
When I first started working on the V PR, I was confused on why the CI didn't run on my branch. The main only setup is unusual and not exactly optimal in my opinion. It forces people to either do work only on the main branch, make a PR to the main branch in this repo or manually run the workflows everytime.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,15 +1,7 @@
 name: CI
 
 # Controls when the action will run.
-on:
-  # Triggers the workflow on push or pull request events but only for the main branch
-  push:
-    branches: [ main ]
-  pull_request:
-    branches: [ main ]
-
-  # Allows you to run this workflow manually from the Actions tab
-  workflow_dispatch:
+on: [push, pull_request, workflow_dispatch]
 
 jobs:
   build:


adsharma:
There may be another reason why the CI didn't run on the V PR. It's because it was your first PR (which requires explicit approval). Subsequent PRs should automatically run CI.

div72:
I was talking about running on the fork, not the main repo. CI does not run on the branches outside of main in this repo either.

adsharma (approved):

adsharma (changes requested):
I thought about it a bit more. Sometimes I push to a feature branch and then create a PR.
I may not want CI to run until the code is ready, at which point I'll create a PR against main.

Looks like the current flow is working for you more or less?

div72:
> Looks like the current flow is working for you more or less?

I am used to it, but still it's not exactly ideal in my opinion.

> Sometimes I push to a feature branch and then create a PR.
I may not want CI to run until the code is ready, at which point I'll create a PR against main.

You can develop it locally and push it only when you want to make a PR, or use `[skip ci]` to prevent a CI from running in a commit.

PR #433, opened 2021-07-14
v: properly pass mutable args

div72:


pyv/transpiler.py
```diff
@@ -222,8 +222,15 @@ def visit_Call(self, node: ast.Call) -> str:
 
         vargs: List[str] = []
 
-        if node.args:
-            vargs.extend(map(self.visit, node.args))
```
adsharma, line 226: I noticed that pyv doesn't have `visit_arg`. Wouldn't it be better to add one and use logic similar to pyrs to handle mutable args?

pyv/transpiler.py
```diff
@@ -222,8 +222,15 @@ def visit_Call(self, node: ast.Call) -> str:
...
+        for idx, arg in enumerate(node.args):
+            if (
+                isinstance(arg, ast.Name)
+                and isinstance(fndef, ast.FunctionDef)
+                and is_mutable(fndef.scopes, fndef.args.args[idx].arg)
+            ):
+                vargs.append(f"mut {arg.id}")
```
adsharma, line 231: `get_id(arg)` is safer.
Also `hasattr(fndef, "args")`.

pyv/transpiler.py
adsharma: @div72 - pinging this thread to make sure you saw my comment.

pyv/transpiler.py
div72: get_id makes more sense, but is there any case in which fndef does not have args? According to the latest Python AST documentation, it seems to be not optional.

pyv/transpiler.py
div72: Correct me if I am mistaken but call args are not ast.arg nodes, no? And in function definitions, making all unknown types generic T is not the best idea as there's no guarantee that the arguments have the same generic type. That's the reason I rewrited it into `function_signature` function.

div72: I guess I can make it back into a visit_arg with holding generic count as an int in the class but that'll get messy when someone uses generics from python side.

adsharma: correct. There is no ast.arg node on Call. Only on FunctionDef.

But if you look at https://github.com/adsharma/py2many/blob/main/pycpp/transpiler.py#L185-L188

The case where different arguments have different generic types can be handled ok.

Having the same general structure as pycpp and pyrs helps promote common code to base class sometimes.

pyv/transpiler.py
adsharma: I think it breaks when using lambdas. The check was introduced in pyrs via 6861cca54e0ae95f194a7d00af43d98b38a7b97b
I just tested that removing the check and running

t -k "rust and ambda" -v

causes a test breakage.

pyv/transpiler.py
div72: Fixed.

PR #434, opened 2021-07-14
v: fix min/max

div72:


pyv/plugins.py
```diff
@@ -44,6 +44,17 @@ def visit_bool(self, node: ast.Call, vargs: List[str]) -> str:
         else:
             return f"({self.visit(node.args[0])}).bool()"
 
+    def visit_min_max(self, node: ast.Call, vargs: List[str]) -> str:
+        self._usings.add("arrays")
+        func = "min" if node.func.id == "min" else "max"
+        # Since python allows calling min/max on either a container or
+        # multiple integers, we simply it to only a container when passing
```
adsharma, line 51: we simply *cast* it to a container?

adsharma (approved):

pyv/plugins.py
div72: Fixed, hopefully the comment makes more sense now.

PR #436, opened 2021-07-19
v: pin vc to fix CI issue

div72:
It seems that while pinning V version to a specific commit, I have not considered the fact that the V's Makefile always fetching latest v.c could break bootstrapping. This PR addresses that by also pinning the version of v.c.

I know this is not ideal, I'll work on getting up-to-date, pre-built builds but in the mean time the CI should not fail for others.


adsharma:
@div72  reverted because it broke CI on ubuntu.

The log was previously available from here

https://github.com/adsharma/py2many/pull/437/checks?check_run_id=3109697454

PR #439, opened 2021-07-25
v: fix a few more test cases

div72:


pyv/transpiler.py
```diff
@@ -25,6 +25,17 @@ def is_mutable(scopes, target: str) -> bool:
     return _is_mutable(scopes, target)
 
 
+def new_expr(source: str, node: ast.AST) -> ast.AST:
```
adsharma: Could move to `py2many/ast_helpers.py` which has similar code.

adsharma (commented):
Looks good. Just one nit and a test failure that needs to be fixed.

pyv/transpiler.py
div72: I used create_ast_node in 301e11b4c04ad80b27e36e4c0d3ddd87a7b36aa2 instead.

div72:
I am actually not sure why the CI is failing. 66d46b73ef4718daa5d202483a0c744ccf05a2b2 should've fixed it and it doesn't fail locally. Need to investigate more.

tests/expected/bitops.v
```diff
@@ -0,0 +1,22 @@
+module main
+
+fn main_func() {
+	mut ands := []bool{}
+	mut ors := []bool{}
+	mut xors := []bool{}
+	for a in [false, true] {
+		for b in [false, true] {
+			ands << ((a && b))
```
adsharma, line 9: Tests are failing because you have:

```
    ast.BitAnd: "&",
```

in pyv/clike.py:139. Does changing this to "&" fix it?

div72: @adsharma I am afraid no. V's bool types do not accept bit-wise ops. if you check below in clike.py(shown below) you could see that there's a special check that changes & to && if any operand is a bool(the test was outputted from my local py2v after all). With a few prints I can see that the target inference doesn't work in CI for some reason but works locally. I will investigate further.
https://github.com/adsharma/py2many/blob/37b42d76cd3b36a78af409b947c438a80ef48f97/pyv/clike.py#L127-L128

adsharma: It fails for me locally as well. Your diagnosis is right - left_type and right_type are None. How did you expect target inference to work locally for you?

div72: Both should've been `bool`s. What's your python version?

div72: Yep, installed python3.8 and it fails for me locally too. Will check what's different in 3.9.

div72: Found the problem, slices in python3.8 is wrapped with a ast.Index class, while in python3.9 are not. Will fix.

div72:
@adsharma Is demorgan test supposed to be also working for other backends?

adsharma:
Nope - demorgan test doesn't make sense for backends other than pysmt.

pyv/clike.py
```diff
@@ -114,8 +114,8 @@ def visit_BinOp(self, node: ast.BinOp) -> str:
         op: str = self.visit(node.op)
         right: str = self.visit(node.right)
 
-        left_type: str = self._typename_from_annotation(node.left)
-        right_type: str = self._typename_from_annotation(node.right)
+        left_type: str = get_inferred_v_type(node.left)
```
adsharma, line 117: `self._typename_from_annotation` has additional magic to handle things like `Dict[Dict[int, str], float]` etc. What problem is this line fixing?

py2many/inference.py
```diff
@@ -521,3 +521,22 @@ def visit_Subscript(self, node):
...
+    def visit_For(self, node):
+        self.visit(node.target)
+        self.visit(node.iter)
+        if (
+            isinstance(node.target, (ast.Name, ast.Attribute))
+            and hasattr(node.iter, "annotation")
+            and isinstance(node.iter.annotation, ast.Subscript)
+        ):
+            typ = node.iter.annotation.slice
+            # ast.Subscript.slice might be wrapped in python < 3.9 with an ast.Index object,
```
adsharma, line 534: Does `self._clike._slice_value()` work here, so we have this logic in one place?

adsharma (commented):
Tests are passing. Yay!

Couple of minor nits.

py2many/inference.py
div72: Fixed.

pyv/clike.py
div72: Reverted, I tried that initially before tweaking with the inference.

div72:
Seems like the cancel previous actions action is broken, my earlier force-push is still running in the CI.

Issue #161, opened 2021-05-04
C++ binit.cpp should use references


adsharma:
The vectors are passed by value. Need to copy the is_reference() logic from rust.

PR #441, opened 2021-08-06
cpp: pass containers by reference

adsharma:
Fixes https://github.com/adsharma/py2many/issues/161

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,7 +1,15 @@
 name: CI
 
 # Controls when the action will run.
-on: [push, pull_request, workflow_dispatch]
+on:
+  # Triggers the workflow on push or pull request events but only for the main branch
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
 
 jobs:
   build:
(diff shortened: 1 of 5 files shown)


pycpp/transpiler.py
```diff
@@ -383,6 +383,11 @@ def visit_arg(self, node):
         typename = "T"
         if node.annotation:
             typename = self._typename_from_annotation(node)
+            # TODO: Should we make this if not primitive instead of checking
+            # for container types? That way we cover user defined structs too.
+            if hasattr(node, "container_type"):
+                # Python passes by reference by default. Rust needs explicit borrowing
```
jayvdb, line 389: "Rust" in pycpp

.github/workflows/main.yml
```diff
@@ -1,7 +1,15 @@
...
-on: [push, pull_request, workflow_dispatch]
+on:
+  # Triggers the workflow on push or pull request events but only for the main branch
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
```
div72, line 12: What about adding something like `if: github.repository != ‘adsharma/py2many’ || github.event.base.ref == 'main'` instead to the jobs?

adsharma: That seems workable. Do you want to submit a new PR? I'll test it this time :)

PR #457, opened 2022-02-04
v: bump version and add initial class support

div72:
Bumps V version to the latest.

~Needs vlang/v#13363 or a PR with the same effect, so I am leaving this as a draft until then.~
Will be unavailable until Monday's release, but should be safe to merge after that.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -66,7 +66,7 @@ jobs:
         if: ${{ !env.WHICH_V }}
         shell: bash
         env:
-          RELEASE: weekly.2022.06
+          RELEASE: weekly.2022.13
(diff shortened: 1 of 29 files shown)


jayvdb:
@div72 , when I run this, I get `error: unknown module attribute `[translated]`.  I tried using `weekly.2022.06` and locally on macos with brew's `0.2.4`

div72:
@jayvdb Are you sure you're invoking the correct V? Any release after vlang/v#13373 should have it.

I'll rebase and fix this today.

div72:
@jayvdb The CI is green, try updating to 2022.13 and ping me if you are still getting errors.

PR #466, opened 2022-04-11
main.yml: Improve dependencies

jayvdb:


jayvdb:
warp4j no longer needed due to use of jgo

PR #468, opened 2022-05-22
Use flutter 2.10.5

jayvdb:
Closes https://github.com/py2many/py2many/issues/467


jayvdb:
Also an into https://github.com/holgerbrandl/homebrew-tap/issues/1

jayvdb:
Also https://github.com/holgerbrandl/kscript/issues/358 , so kotlin on macos is broken until that is fixed

PR #469, opened 2022-05-22
Add recursive vlang test

jayvdb:


tests/test_transpile_self.py
```diff
@@ -246,3 +257,29 @@ def test_julia_recursive(self):
...
+        assert_counts(
+            *_process_dir(
+                settings,
+                PY2MANY_MODULE,
+                OUT_DIR,
+                False,
+                _suppress_exceptions=(AstNotImplementedError, AstUnrecognisedBinOp),
+            ),
+            format_error_count=2,
+            failure_count=12,
```
jayvdb, line 284: ping @div72 , it would be good to get these errors down to 0; probably only a few bits of syntax needed.

div72: I'll take a look.

jayvdb: Created https://github.com/py2many/py2many/issues/474 and gave it a `V` label

PR #477, opened 2022-05-30
main.yml: Use linuxbrew

jayvdb:
This is exploratory , hoping that maybe we can also get linuxbrew working on WSL, in order to bring linter/formatter tools to all three platforms via a common package manager, without any reliance on fetching specific versions.


jayvdb:
Im less likely to try to complete this now that we've got jgo looking after the Java based tools

PR #483, opened 2022-06-07
pyjl: Add exception handling

jayvdb:


div72:
@jayvdb On the topic, do you have anything on mind for handling exceptions on languages like Go/Rust/V where error handling is dissimilar to Python?

jayvdb:
@div72 , unfortunately that nut has been put into the too-hard-basket many times.  I've created https://github.com/py2many/py2many/issues/484 to pool our resources on it.

PR #487, opened 2022-06-10
Improve declarations_with_defaults

jayvdb:


MiguelMarcelino:
I think we can close this. I must have a better look at what this is and how we could easily integrate it.

jayvdb:
I'll take a closer look at this one, but I am inclined to agree with you that this one has only marginal benefits so may not be worth keeping around as-is.

PR #492, opened 2022-06-13
Support casting ctypes

jayvdb:


Askr46 (approved):

adsharma:
The PR is more than a year old and needs rebase. In the meanwhile you can use:

```
from ctypes import c_int8 as i8, c_int16 as i16, c_int32 as i32, c_int64 as i64
from ctypes import c_uint8 as u8, c_uint16 as u16, c_uint32 as u32, c_uint64 as u64

def main():
    a: u16 = 1
    print(a)
    b: i64 = 1
    print(b)
    c: u64 = 1
    print(c)

if __name__ == "__main__":
    main()
 ```

PR #493, opened 2022-06-14
Julia fix Range

MiguelMarcelino:
Ranges have to be translated to the Julia language, as Julia uses 1-based indexing, whereas Python uses 0-based indexing. This fix should


tests/cases/math_func.py
```diff
@@ -1,5 +1,5 @@
 def main():
-    a = 2**4
+    a = 2 ** 4
```
jayvdb, line 2: this needs to be reverted

.gitignore
```diff
@@ -25,3 +25,7 @@ tests/cases/*.rs
 tests/cases/*.cpp
 tests/cases/*.kt
 tests/cases/*.nim
+dist/
+pyjl/test_files
+test.sh
+tests/output
```
jayvdb, line 31: `pyjl/test_files` and `test.sh` dont belong in here (yet?).  

`tests/output` is useful, but could you add an end-of-line (EOL) at end-of-file (EOF) here. github and other tools show missing EOL as an error

pyjl/plugins.py
```diff
@@ -9,7 +9,8 @@
 
 try:
     from argparse_dataclass import dataclass as ap_dataclass
-    from argparse_dataclass import ArgumentParser
+
+    # from argparse_dataclass import ArgumentParser
```
jayvdb, line 13: why is this commented out?

tests/cases/math_func.py
jayvdb: or maybe this is ok, but `tests/expected/math_func.py` needs to be updated - one of these is causing CI to fail.

jayvdb: I've now checked, and there is no reason for this file or `tests/expected/math_func.py` to be updated.  Please restore them to the same as on the `main` branch.

tests/cases/math_func.py
```diff
@@ -4,4 +4,4 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    main()
```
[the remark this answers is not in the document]
MiguelMarcelino: The one currently on main does not have an EOL.

.github/workflows/main.yml
```diff
@@ -19,7 +19,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [ '3.8' ]
+        python-version: [ '3.9' ]
```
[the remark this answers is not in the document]
MiguelMarcelino: I believe that `py2many/ast_helpers` uses `ast.unparse`, which was only introduced in Python 3.9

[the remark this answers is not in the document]
jayvdb: We have a fallback in that file to use `from astor import to_source as unparse` on py38.  Is that not working for you?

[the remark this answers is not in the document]
MiguelMarcelino: I believe this error happened when I tried to test with `pytest-3`, but in GitHub's workflows it is fine. Anyway, I believe I also use some features from Python 3.9 in the PyJL transpiler, that is why I ultimately left it in.

tests/expected/math_func.py
```diff
@@ -10,4 +10,4 @@ def main_func():
 
 
 if __name__ == "__main__":
-    main_func()
+    main_func()
```
jayvdb, line 13: also restore this EOL at EOF

tests/expected/math_func.jl
```diff
@@ -7,4 +7,4 @@ function main()
     main_func()
 end
 
-main()
+main()
```
jayvdb, line 10: an EOL is missing here too

PR #495, opened 2022-06-14
Julia main rewriter

MiguelMarcelino:
Rewrites `if __name__ == "__main__": ...` to Julia. 
**Possible issues**: Due to the way this affects scoping, this change requires removing type hints from Python's `main` method. As an example, consider the test `tests/cases/sys.argv.py`, which uses type hints in the main method. As the `if` node does not introduce a new scope in Julia and type declarations are not supported in global variables (at least in Julia 1.7.3), this will not be allowed.


jayvdb:
As this removes `tests/expected/sys_argv.jl` , and doesnt add any new features, I think this isnt ready yet.  Being able to support command line arguments is a critical feature.

MiguelMarcelino:
Closing, as it did not add any value

PR #498, opened 2022-06-19
Add float() on Dart, Go, Kotlin, Nim and V

### LINKED ISSUE #129: float(1) fails on Dart, Go, Kotlin and Rust
float() is not a defined function for these languages.

jayvdb:
Fixes #129

### DIFF
--- a/pyv/inference.py
+++ b/pyv/inference.py
@@ -11,7 +11,7 @@ from py2many.inference import get_inferred_type, InferTypesTransformer
 
 V_TYPE_MAP: Dict[type, str] = {
     int: "int",
-    float: "f32",
+    float: "f64",
     bytes: "[]byte",
     str: "string",
     bool: "bool",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ strum_macros = "*"
 lazy_static = "*"
 pylib = "*"
 tempfile = "*"
+float-ord = "*"
 
 
 [lib]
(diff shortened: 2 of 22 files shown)


div72:
This might fail for non-numeric casts in V as .f32()/.f64() methods are used instead for other types. Would using the inferred type or falling back to a helper method when type cannot be inferred to offload the work to V compiler be the best action in this case?

jayvdb:
@div72 , can you suggest a test case which would fail with this PR?   Especially if the test case fails on V only, but passes the others?  i.e. I'm trying to determine if there is a V specific weakness in this PR.

pyv/inference.py
```diff
@@ -11,7 +11,7 @@
 
 V_TYPE_MAP: Dict[type, str] = {
     int: "int",
-    float: "f32",
+    float: "f64",
```
jayvdb, line 14: @div72 , is this change going to cause you any problems?  `f64` is closer to what `float` is in Python, but I dont want to disrupt your use of py2many if it can be avoided, and iirc this change isnt needed for the `float(1)` test to pass.

jayvdb:
btw we now have a Discord server at https://discord.gg/xWrjYf32 if you want to join in; very low traffic.

div72:
`float("2.71")` should fail on V, not sure about others.

-----

I don't use Discord but I can chime in if you're willing to use a Matrix bridge.

pyv/inference.py
div72: Nope, feel free to. I decided to go with f32 to lessen the size of structs IIRC.

jayvdb:
Ah, `float("2.71")` .. fails for all targets, except for Kotlin which converts it to `"2.71".toDouble()`.  That is likely to remain not supported for a long time, unless someone wants to get it working for most targets.

PR #502, opened 2022-06-20
v: initial exception support

div72:


div72:
Finally statements are not properly implemented, defer in V is function based and not block based and I'm not sure on how to implement it otherwise without adding bunch of gotos.

Exception hierarchies are also not implemented, V does not sub-classes so also not sure on how that'd function.

Lastly tests/cases/exception_names.py is a bit off an oddball. vexc adds a signal handler, but it raises ArithmeticError on SIGFPE but the script catches ZeroDivisionError. Another point is that 3 / 0 is a compile error for V, which means the script won't even compile in the first place.

div72:
@jayvdb [Getting a timeout for setup-java action](https://github.com/py2many/py2many/runs/6960891290?check_suite_focus=true), restarting.

jayvdb (approved):

jayvdb:
@div72 , If you want, you can rebase and then I can do a fast forward merge, or I can do a squash and merge which adds me as the committer for no reason.  Once you let me know your preference, I'll keep following whichever approach you prefer.

div72:
@jayvdb I don't really mind if you become the committer, ping me on GitHub(a message would be faster but I don't use Discord and there's no Matrix bridge) and merge as you see fit if I don't respond within 5 minutes.

div72:
@jayvdb I am not familiar with the setup-vlang action is the V version not pinned? Nevertheless I'll fix it in a bit.

jayvdb:
The V version isnt pinned in our `main.yml` - if it is, it wouldnt match the vlang that people will download.

If you cant fix it easily, https://github.com/vlang/setup-v does allow pinning the version, so we can do that as a temporary measure.

div72:
@jayvdb A bit delayed but fixed now. I'll add periodic CI so that something like this can be catched faster.

PR #519, opened 2022-07-29
Fix translation of for-else and while-else statements

MiguelMarcelino:
This is a proposal to translate `for-else` and `while-else` loops from Python.
This relates to this [issue](https://github.com/py2many/py2many/issues/415) from Py2Many.


MiguelMarcelino:
There appears to be a weird linker bug when compiling with g++ for macOS. Perhaps it is related to [this discussion](https://github.com/Homebrew/discussions/discussions/3659).

jayvdb:
I believe this was the same as https://github.com/py2many/py2many/pull/530

PR #521, opened 2022-07-30
pycpp: misc fixes

### LINKED ISSUE #520: py2many.exceptions.AstNotImplementedError: sequence item 0: expected str instance, NoneType found
Hi,

I'm invoking:

```
09:44:42 1d folkert@belle:~/git/bf-compiler master(2)* ± py2many --cpp=1 mandelbrot.py --outdir t/
Writing to: t
mandelbrot.py ... t/mandelbrot.cpp
mandelbrot.py:14:0: py2many.exceptions.AstNotImplementedError: sequence item 0: expected str instance, NoneType found
```

[mandelbrot-python.zip](https://github.com/py2many/py2many/files/9225635/mandelbrot-python.zip)

div72:
Fixes #520.

### DIFF
--- a/pycpp/plugins.py
+++ b/pycpp/plugins.py
@@ -10,6 +10,8 @@ import time
 from tempfile import NamedTemporaryFile
 from typing import Callable, Dict, List, Tuple, Union
 
+from py2many.analysis import get_id
+
 try:
     from argparse_dataclass import dataclass as ap_dataclass
     from argparse_dataclass import ArgumentParser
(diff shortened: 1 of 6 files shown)


pycpp/plugins.py
```diff
@@ -52,13 +55,27 @@ def visit_named_temp_file(self, node, vargs):
         node.result_type = True
         return "NamedTempFile::new()"
 
+    def _translate_file(file: str) -> str:
+        FILE_MAP: Dict[object, str] = {
+            sys.stdout: "std::cout",
+            sys.stdin: "std::cin",
+            sys.stderr: "std::cerr",
+        }
+        return FILE_MAP.get(class_for_typename(file, file), file)
```
div72, line 64: Should class_for_Typename return sys.stdout/stdin/stderr or io.TextIOWrapper here?

div72: @adsharma @jayvdb

jayvdb: IMO they should be `io.TextIOWrapper`.

We already have plugins for `io.TextIOWrapper.read` and `io.TextIOWrapper.write`

For this PR a new test case is needed in tests/cases/ - maybe called `stdio.py`.  The test framework cant handle stdin, but a test case containing `sys.stdout.write("foobar\n")` should suffice to validate that the changes in this PR are working and cant regress.

adsharma:
Could you please rebase and get the tests to pass?

PS: I managed to find a few hours to check on py2many this weekend. Not too sure how much time I'll be able to spend. Let's try to commit as many of these important fixes already in the queue as possible.

div72:
@adsharma MacOS CI failure seems unrelated: https://stackoverflow.com/questions/73714336/xcode-update-to-version-2395-ld-compile-problem-occurs-computedatomcount-m

adsharma:
Looks good to me. The only thing I don't understand is the special casing of `sys` here:

```

div72:
@adsharma Not sure how much of a problem this is, but sys.exit does have `__self__` which causes that piece of code to get triggered when it shouldn't.

jayvdb (approved):

PR #528, opened 2022-10-08
Add missing generic_visit()

adsharma:
Without this, some assignment statements are not properly annotated with variable context (e.g. which variables are assigned inside the body).

We were hitting this case after a rewrite.

Fixes: https://github.com/py2many/py2many/issues/527

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -105,6 +105,11 @@ test = true
 path = "tests/expected/hello_world.rs"
 
 [[bin]]
+name = "ifexp"
+test = true
+path = "tests/expected/ifexp.rs"
+
+[[bin]]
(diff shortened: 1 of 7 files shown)


jayvdb:
Can we get a test associated with this change? i.e. adding an assignment statement which was previously not being annotated correctly.

adsharma:
I'm thinking some of the long running tests are valuable, but should not be commit blocking.
How can we get to a faster turn around with high signal?

jayvdb (approved):

jayvdb:
> How can we get to a faster turn around with high signal?

https://github.com/py2many/py2many/issues/540 should be quite easy, and have reasonable results.

PR #532, opened 2022-10-11
PyJL dispatch improvements and increased `stdlib` support

MiguelMarcelino:
Improvements to the current dispatch mechanism, including the removal of string-based fallbacks. 
A wider range of Python `stdlib` functions is now supported.


MiguelMarcelino:
This PR is too large for review. Closing

PR #533, opened 2022-10-13
Add support for external configuration files

MiguelMarcelino:


.github/workflows/main.yml
```diff
@@ -131,7 +132,9 @@ jobs:
                   kscript \
                   ktlint \
                   z3 \
+                  libyaml \
                 && true
+              python -m easy_install pyyaml
```
jayvdb, line 137: setup.py `install_requires` is where runtime dependencies need to be added.

py2many/input_configurations.py
```diff
@@ -0,0 +1,231 @@
...
+    input = Path(filename)
+    if not input.is_file():
+        raise Exception("The input configuration file does not exist")
+    if file_extension == ".ini":
+        return ConfigFileHandler(input)
+    else:
+        raise Exception("Configuration file has to be a .ini file")
+
+
+class ConfigFileHandler:
```
jayvdb, line 28: we need a test to cover this.

jayvdb (changes requested):

PR #539, opened 2023-02-12
Use jgo for Java based formatters

jayvdb:
Fixes https://github.com/py2many/py2many/issues/509

Uses https://github.com/scijava/jgo to remove the need for setting up ktlint and cljstyle , which especially removes the need to keep the versions in sync.

Ideally kscript is also run under jgo, but that is blocked by https://github.com/scijava/jgo/issues/89

Another approach is to use another kotlin scripting runner, e.g. https://stackoverflow.com/questions/34974039/how-can-i-run-kotlin-script-kts-files-from-within-kotlin-java/75424410, but I havent found one which is a drop in solution. e.g. https://github.com/mpetuska/ktx/issues/28 and https://github.com/scijava/scijava-common/issues/451

So far https://github.com/scijava/scijava-common/issues/451 looks easiest to workaround.


jayvdb:
Raised https://github.com/scijava/jgo/issues/90 about current failure

tox.ini
```diff
@@ -24,6 +24,7 @@ deps =
     toposort
     argparse_dataclass
     git+https://github.com/adsharma/adt/
+    git+https://github.com/jayvdb/jgo@sort-jars
```
jayvdb, line 27: this is https://github.com/scijava/jgo/pull/97

PR #553, opened 2023-08-07
v: add support for string multiplication

div72:


div72:
@jayvdb conan is failing in the CI.

MiguelMarcelino:
@div72 I believe I made it work in PR #554. I am still checking if tests run properly, but I believe we could merge the changes if the fix works properly and if @jayvdb agrees.

jayvdb:
The conan problem currently being fixed in https://github.com/py2many/py2many/pull/558

There is a vlang problem there - looks like a simple update to the output needed, but I dont have vlang available on openSUSE tumbleweed, so I cant update that way, but I'll try to get https://build.opensuse.org/package/show/home:MargueriteSu/vlang building, or manually make the changes that CI is expecting.

PR #554, opened 2023-12-24
feat: Add Julia rewriter for boolean operations

MiguelMarcelino:
Adding a rewriter for Julia's boolean operators. This should convert any Python comparisons that are incompatible in Julia.

TODO:
- [x] Transpile the operations to Julia-compatible ones
- [ ] Run tests and ensure no behaviour is broken

NOTE: Merge only after #555 is merged.


MiguelMarcelino:
Closing in favour of #556

PR #556, opened 2023-12-26
pyjl: Translate boolean operations

MiguelMarcelino:
Adding a rewriter for Julia's boolean operators. This should convert any Python comparisons that are incompatible in Julia.

TODO:
- [x] Transpile the operations to Julia-compatible ones
- [x] Run tests and ensure no behaviour is broken


jayvdb:
Looking good.  There were a few failures locally when I tested it.

MiguelMarcelino:
Many thanks! Will have another look and hopefully fix things during this next week.

jayvdb:
Nice. The Julia cases are now all passing. I can tackle the 4 failures in non-Julia tests if you'd prefer.

jayvdb:
I see two of the same errors on https://github.com/py2many/py2many/pull/566 , so I will create a task about that and fix it shortly.

jayvdb:
This is now failing only because it doesnt reach 90% test coverage.
Specifically less than 50% of the new `helpers.py`, so much of it isnt being used yet
```
helpers.py                     40     22    45%
```

jayvdb:
I've reduced the coverage target to allow this to pass, so that we can focus on getting https://github.com/py2many/py2many/pull/571 passing and merged.

PR #562, opened 2023-12-31
Fix typpete imports

jayvdb:


tests/test_inference.py
```diff
@@ -0,0 +1,32 @@
...
+        infer_types(tree)
+        assert hasattr(tree.body[0].targets[0], "annotation")
+        assert tree.body[0].targets[0].annotation.value.id == "List"
+        assert tree.body[0].targets[0].annotation.slice.id == "int"
+
+
+class TestInferenceTyppete(unittest.TestCase):
+    def test_infer_types_list_failure(self):
+        tree = parse("a = [10, 20]")
+        with self.assertRaises(AttributeError):
```
jayvdb, line 24: This is caused by lack of Python 3.9+ support in Typpete, c.f. https://github.com/adsharma/Typpete/pull/1

PR #571, opened 2024-01-03
Feature/translate unit tests

MiguelMarcelino:
This PR offers two unittest functionalities that are very important for translations to ensure the semantics are identical:
1. A generic unit-test translation approach translates a subset of Python's `unittest` framework and is compatible with all the supported languages,
2. Parameterized unit tests for PyJL, as these largely benefit the reuse of test methods. Python has support for this functionality thought the `pytest` testing framework. This PR offers a translation of a subset of this module to Julia.

To be merged after #556.


pyjl/juliaAst.py
jayvdb, line 1: rename to `julia_ast.py` ;-)

jayvdb:
I'm working on the CI failures at https://github.com/py2many/py2many/pull/580

pyjl/transpiler.py
```diff
@@ -18,6 +18,15 @@
     SMALL_USINGS_MAP,
 )
 
+from py2many.analysis import get_id, is_void_function
+from py2many.declaration_extractor import DeclarationExtractor
+from py2many.clike import _AUTO_INVOKED, class_for_typename
+from py2many.tracer import is_list, defined_before, is_class_or_module, is_enum
+
+from pyjl import julia_ast
```
jayvdb, line 26: When we install py2many, the transpilers are put under py2many.
To allow for that, change this to
```py
try:
    from py2many.pyjl import julia_ast
except ImportError:
    from pyjl import julia_ast
```

jayvdb:
maybe disable py38 in https://github.com/py2many/py2many/blob/c3bf3e4/.github/workflows/main.yml#L28 ?
Then we can get py38 working later, or even decide whether we want to keep py38 support.

tests/cases/unittest.py
jayvdb, line 1: when I run this test locally, I see
```py
Traceback (most recent call last):
  File "/home/jayvdb/python/py2many/tests/test_cli.py", line 223, in test_generated
    expected_output = get_python_case_output(
  File "/home/jayvdb/python/py2many/tests/test_cli.py", line 146, in get_python_case_output
    raise RuntimeError(f"Invalid {case_filename}:\n{proc.stdout}{proc.stderr}")
RuntimeError: Invalid /home/jayvdb/python/py2many/tests/cases/unittest.py:
b''b'Traceback (most recent call last):\n  File "/home/jayvdb/python/py2many/tests/cases/unittest.py", line 1, in <module>\n    import unittest\n  File "/home/jayvdb/python/py2many/tests/cases/unittest.py", line 3, in <module>\n    class UnitTestTranslation(unittest.TestCase):\nAttributeError: partially initialized module \'unittest\' has no attribute \'TestCase\' (most likely due to a circular import)\n'
```

We had the same problem with a test case `math.py`.

The solution is to rename the test to not match a Python stdlib name.
e.g. 'unit_test'

tests/cases/unittest.py
```diff
@@ -0,0 +1,15 @@
...
+        z = x  # Added
+
+        self.assertEqual(z, [1, 2])
+
+        x = [1, 2, 3]
+        y = x
+        self.assertTrue(x is y)
+
+if __name__ =="__main__":
+    unittest.main()
```
jayvdb, line 15: when I run this locally, the test case fails with

```py
Traceback (most recent call last):
  File "/home/jayvdb/python/py2many/tests/test_cli.py", line 226, in test_generated
    self.assertTrue(expected_output, "Test cases must print something")
  File "/usr/lib64/python3.9/unittest/case.py", line 688, in assertTrue
    raise self.failureException(msg)
AssertionError: b'' is not true : Test cases must print something
```

This restriction is in place because `print` is available in all the transpilers.

Solution is to replace this line with
```py
if __name__ =="__main__":
    print("Running tests")
    unittest.main()
```

tests/expected/unittest.jl
```diff
@@ -0,0 +1,20 @@
+import unittest
```
jayvdb, line 1: when I try running the .jl locally I get
```
Traceback (most recent call last):
  File "/home/jayvdb/python/py2many/tests/test_cli.py", line 305, in test_generated
    assert (
AssertionError: Execution of unit_test.jl failed:
b''b'ERROR: LoadError: ArgumentError: Package unittest not found in current path.\n- Run `import Pkg; Pkg.add("unittest")` to install the unittest package.\nStacktrace:\n [1] macro expansion\n   @ ./loading.jl:1630 [inlined]\n [2] macro expansion\n   @ ./lock.jl:267 [inlined]\n [3] require(into::Module, mod::Symbol)\n   @ Base ./loading.jl:1611\nin expression starting at /home/jayvdb/python/py2many/tests/build/unit_test.jl:1\n'
```
However I cant find a Julia Package called unittest, or much Julia code on Github which uses ["assertEqual"](https://github.com/search?q=assertEqual+language%3Ajulia&type=code) .

MiguelMarcelino: This would require a special translation that I have implemented in my fork. But I will maybe try to split it up to avoid introducing too many changes at once.

PR #597, opened 2024-07-06
Assess a more streamlined CI/CD model

adsharma:
The idea is to get a clear signal on a PR within 5 minutes.

We could still run the more comprehensive CI/CD periodically and open bug reports for things that are broken.


adsharma:
The shorter CI runs in 2m 46 secs
I tested a repeat run (cached venv) to run in 2m 41 secs

jayvdb (approved):

Issue #600, opened 2024-07-07
Go test failures


adsharma:
Failure signature:

```
Writing to: /home/build/py2many/tests/build
loop.go:5:2: no required module provides package github.com/hgfischer/go-iter: go.mod file not found in current directory or any parent directory; see 'go help modules'
```

adsharma:
This is a docker specific failure caused by missing setup actions.

Previously we discussed: `act`. Instead of repeating ourselves, perhaps we can run act to get the same setup as GHA?
https://nektosact.com/installation/aur.html

PR #601, opened 2024-07-07
Update dockerfile to match our workflows

### LINKED ISSUE #599: C++ test failures
This is reproducible by running tests in the recently updated docker image `archlinux:py2many`

```
Writing to: /home/build/py2many/tests/build
/home/build/py2many/tests/build/comb_sort.cpp:4:10: fatal error: 'cppitertools/range.hpp' file not found
    4 | #include <cppitertools/range.hpp>  // NOLINT(build/include_order)
      |          ^~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

adsharma:
Fixes #599, #600


adsharma:
https://github.com/GoogleContainerTools/distroless is another option for this. They use debian12 or alpine as the base distro.

Sample Dockerfile with venv: https://github.com/psf/black/blob/main/Dockerfile

runtime/go-setup.sh
```diff
@@ -0,0 +1,11 @@
+mkdir -p tests/build
```
jayvdb, line 1: imo this directory should be called "scripts"
runtime has a different meaning

jayvdb (approved):

PR #608, opened 2024-07-15
Add dlang to docker

adsharma:


adsharma:
@mw66 I see 9 test failures with this dockerfile:

```
$ dmd --version
DMD64 D Compiler v2.109.1
```

Many of them are due to a whitespace/formatter issue. A coupe of them seem to be more involved.

```
docker buildx build -t archlinux:dlang --build-arg DLANG=1 . -f docker/Dockerfile
docker run -v $(pwd):/home/build/py2many -it archlinux:dlang
tox -- -k dlang
```

to reproduce.

adsharma:
Also all the `dlang` tests were skipped in the CI [run](https://github.com/py2many/py2many/actions/runs/9945974143/job/27475451066). May be because the toolchain was not installed?

mw66:
I tested locally.

So is this Dockerfile the file to do dmd setup?

https://github.com/py2many/py2many/blob/main/docker/Dockerfile

adsharma:
Yes, this PR already has the `Dockerfile` changes you need. 

However CI doesn't run docker. You'll need something like [this](https://github.com/py2many/py2many/blob/main/.github/workflows/main.yml#L153-L163) `vlang` segment to enable `dlang`.

If the setup is verbose, suggest doing something similar to `scripts/go-setup.sh` so it could be shared between docker and CI.

PR #613, opened 2024-07-17
update dlang in .github/workflows/main.yml

mw66:
update dlang in .github/workflows/main.yml

https://github.com/py2many/py2many/issues/610

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -112,6 +112,9 @@ jobs:
         run: |
           set -ex
           if [ "$RUNNER_OS" == "Linux" ]; then
+              sudo wget https://netcologne.dl.sourceforge.net/project/d-apt/files/d-apt.list -O /etc/apt/sources.list.d/d-apt.list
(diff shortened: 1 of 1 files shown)


adsharma:
The code looks good, but I can't merge it because of the conflicts.
Could you try rebasing like so:

```
git rebase -i origin/main
git push -f
```

mw66:
Done, can you try again?

adsharma:
Any reason not to use [this](https://github.com/dlang-community/setup-dlang) to cover mac and windows?

Also, could you honor `FOCUS` so if someone is trying to run CI for a different language, we don't pay the cost of pulling these packages?

mw66:
Ahh, at this stage I think you know much more than I do on setting up github.

I'm a regular desktop D user, and only know very basic github.

Can you take over this PR from here?

adsharma:
Sure. I'll give it a shot this weekend.
@jayvdb knows more about it than I do :)

PR #614, opened 2024-07-17
change global (const) var to upper case for Dlang

mw66:
Dlang need `const` information to do proper global var initialization.

Python does not really have const var, the typical convention is to use upper case global var as const.

So it's better to use this convetion.

But I don't know how this will affect other language transpilers.


adsharma:
In the variant of python consumed by py2many (docs/langspec.md), these conventions are unnecessary.

There is a compiler pass that figures out if the variable is `const` or not by performing a mutability analysis and annotates the AST node. That info is used to emit the right keyword for each backend.

If the convention in `dlang` is to use upper case, you can do so in the `pyd` backend.

adsharma:
Please reopen a new one if this is still an issue.

PR #625, opened 2024-08-11
fix tests/test_unsupported.py dict_del

mw66:


pyd/transpiler.py
```diff
@@ -473,9 +473,17 @@ def visit_Tuple(self, node) -> str:
...
-        self._usings.add("core.memory")
         target = node.targets[0]
         target_str = self.visit(target)
+        if len(node.targets) == 1 and isinstance(target, ast.Subscript):
+            self._usings.add("std.algorithm")
+            # special handling to delete element from container, e.g.
+            # tests/test_unsupported.py dict_del: "a = {1: 1}; del a[1]; assert not a",
+            value = self.visit(target.value)
+            index = self.visit(target.slice)
+            return "{0}.remove({1});".format(value, index)
```
adsharma, line 484:
```
f"{value}.remove({index})
```

is more readable. But much of py2many was written before f-strings were prevalent. So this is ok for consistency.

PR #629, opened 2024-08-12
Add --version flag to cli

adsharma:
https://packaging.python.org/en/latest/guides/single-sourcing-package-version/


adsharma:
Ignoring the lint error. May be a good reason to switch to `flake8`.

PR #635, opened 2024-09-07
Initial support for mojo

adsharma:
[Mojo](https://www.modular.com/mojo) is a recent language from modular.


adsharma:
Most of the code is copy pasta from pynim. Needs cleanup.

PR #656, opened 2024-10-21
Make _typename_from_annotation a classmethod

adsharma:
There is a long standing code duplication issue:

py$lang/inference.py uses `get_id(node.annotation)` to infer types.
py$lang/transpiler.py uses ` _typename_from_annotation(node)` to infer types.

The latter is more powerful and should be used in `inference.py` as well. However, due to a circular dependency, inference.py can't use `_typename_from_annotation`. One way to resolve it by organizing type inference code into a sequence of classmethods.

This PR does it only for rust. If it passes tests, we could repeat it for all other targets and then fix up `inference.py` to use the more comprehensive method.

### DIFF
--- a/py2many/clike.py
+++ b/py2many/clike.py
@@ -119,16 +119,16 @@ class CLikeTranspiler(ast.NodeVisitor):
     NAME: str
 
     builtin_constants = frozenset(["True", "False"])
+    _default_type = _AUTO
+    _type_map = {}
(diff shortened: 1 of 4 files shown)


adsharma:
This code can be cleaned up as well:

https://github.com/py2many/py2many/blob/main/py2many/inference.py#L176-L177

PR #695, opened 2025-02-15
[AI refactor]: move common code to py2many/inference.py

adsharma:
Most of this refactor was done by an AI with some human prompting.

The commit is in the right direction (reduces copy-pasta). But doesn't pass all the tests yet.

### DIFF
--- a/py2many/inference.py
+++ b/py2many/inference.py
@@ -105,6 +105,50 @@ def is_compatible(
     return True
 
 
+class LanguageInferenceBase:
+    """Base class for language-specific type inference"""
+
+    TYPE_MAP = {}  # Should be overridden by subclasses
+    CONTAINER_TYPE_MAP = {}  # Should be overridden by subclasses
+    WIDTH_RANK = {}  # Should be overridden by subclasses
+
+    @classmethod
+    def map_type(cls, typename):
+        typeclass = class_for_typename(typename, None)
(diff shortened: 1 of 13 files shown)


adsharma:
Baseline: [12 failures](https://github.com/py2many/py2many/actions/runs/13348015020/job/37280913297). 
After: [25 failures](https://github.com/py2many/py2many/actions/runs/13346975004/job/37278629324): 

6 failures from Rust.
7 failures from other languages.

py2many/inference.py
```diff
@@ -105,6 +105,50 @@ def is_compatible(
...
+    TYPE_MAP = {}  # Should be overridden by subclasses
+    CONTAINER_TYPE_MAP = {}  # Should be overridden by subclasses
+    WIDTH_RANK = {}  # Should be overridden by subclasses
+
+    @classmethod
+    def map_type(cls, typename):
+        typeclass = class_for_typename(typename, None)
+        if typeclass in cls.TYPE_MAP:
+            return cls.TYPE_MAP[typeclass]
+        return typename
```
cclauss, line 120:
```suggestion
        return cls.TYPE_MAP.get(typeclass, typename)
```
% [`ruff rule SIM401`](https://docs.astral.sh/ruff/rules/if-else-block-instead-of-dict-get)

PR #698, opened 2025-03-04
Add cpplint to pre-commit

cclauss:
___.pre-commit-config.yaml___
```yaml
-   repo: https://github.com/cpplint/cpplint
    rev: 2.0.0
    hooks:
      - id: cpplint
        args:
          - --filter=-legal/copyright,-whitespace/semicolon,-runtime/reference
          - tests/expected/*.cpp
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,4 +28,5 @@ repos:
     hooks:
       - id: cpplint
         args:
-          - --filter=-legal/copyright,-whitespace/semicolon,-runtime/reference tests/expected/*.cpp
+          - --filter=-legal/copyright,-whitespace/semicolon,-runtime/reference
+          - tests/expected/*.cpp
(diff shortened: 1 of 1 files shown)


adsharma:
How about using https://github.com/pocc/pre-commit-hooks

So we can have both `cpplint` and `clang-tidy`?

```
$ cat .clang-tidy
Checks: "-clang-analyzer-deadcode*"
```

cclauss:
I will try that in a parallel pr but I chose this because pocc was last updated four years ago and its README says its ci tests are failing and because the proposed one is from the cpplint team.

adsharma:
@cclauss I noticed that as well. My concern comes from:

https://www.google.com/search?q=clang-tidy+vs+cpplint+site%3Agithub.com

A number of projects switched to clang-tidy over the last few years for reasons I'm not completely familiar with. I noticed you work on `cpplint` and care about static checking C++. Thank you for the contribution!

The good thing about static checkers is it's not a one or the other game. As long as they're quick we can have both.

Issue #770, opened 2026-05-21
Python AST to <many> AST


IvanGrigorik:
AFAIK from transpiler files (e.g, the one for [cpp](https://github.com/py2many/py2many/blob/453011646b87e57ba006a9ce3f58a79dd4c0d21e/pycpp/transpiler.py)), the py2many will walk through the Python AST and gradually transform each node to, for example, C++ text.
For the rest of this issue / proposal, I will use C++ as an example,  since this is the main language that I am interested in.

However, when Py2many users want to make code modifications, they will be limited by that (e.g., you need to transform the function, load it into a third library to get IR, and modify the third-library-defined IR).

It will be really beneficial and useful to have AST-to-AST or other IR-to-IR transformations. That allows not just code transformation, but also potential optimizations based on the transformed code.

The workflow can look like that
```
------------------------------------
|         [Python src]             |
|              |                   |
|              v                   |
|     py2many.produce_IR(...)      |
|              |                   |
|              v                   |
|         [Python IR]              | Full Python code
|              |                   |
|              v                   |
| py2many.translate_IR_to_cpp(...) |
|              |                   |
|              v                   |
|       py2many.serialize()        |
|              |                   |
---------------|--------------------
               v
           [CPP src]                  C++ src result

```
That allows users to do code transformation step-by-step.

@adsharma, since you are the most active developer of this tool, may you address this question and check if it is feasible?

adsharma:
@IvanGrigorik - this would be a stronger design than the one we have. So why didn't we do it this way?

- py2many was forked from pyrs and py14 before that. They used the same visitor pattern (AST -> str)
- for the AST -> AST transformation to work, you'll need a python AST definition for every target language, which didn't exist back then
- treesitter was released circa 2018 I think. It seems to have many parsers and AST definitions accessible via python

Worth pursuing.

IvanGrigorik:
@adsharma https://github.com/py2many/py2many/pull/771 looks great. This is something that we would like to see in this project. 
I hope that py2many will support more languages with this translation pipeline.
Thanks!

adsharma:
@IvanGrigorik added something similar for C++. After the PR lands this issue will be closed. You're welcome to submit further improvements to complete the transition from str -> AST node.

PR #771, opened 2026-05-21
Add Rust AST render pass

adsharma:
Related: #770 

Already converted:

  - module/source file wrapper
  - simple calls
  - returns
  - assignments / let
  - augmented assignments
  - expression statements
  - functions
  - for, while, loop, normal if
  - asserts / macro calls
  - await / yield
  - print / drop

  Still mostly string-based:

  - expressions: binary ops, unary ops, comparisons, lambdas, attributes, subscripts, slices, tuples
  - literals/containers: lists, dicts, sets, bytes
  - structs/classes/enums/sealed classes
  - imports/usings/features/prologue
  - global lazy static collection initialization
  - generator/list comprehensions
  - try/unsupported exception output
  - extension module generation
  - dispatch/plugin functions still return strings
  - inherited py2many.clike visitors still build strings internally, then get wrapped as RustNode

  Also worth improving:

  - Replace the `__getattribute__` visitor wrapper with explicit node returns once enough visitors are converted. It works, but it is transitional and adds overhead.
  - Add stricter tests that direct visitor calls return specific node classes, not just RustNode.
  - Add a small renderer test per construct to prevent accidental string-only regressions.
  - Eventually stop subclassing str; use proper AST classes and render only at the final pass. Right now RustNode(str) is a compatibility bridge.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,6 +20,8 @@ classifiers = [
 ]
 dependencies = [
     "argparse_dataclass",
+    "tree-sitter",
+    "tree-sitter-rust",
 ]
 dynamic = ["version"]
 
--- a/pyrs/clike.py
+++ b/pyrs/clike.py
@@ -10,6 +10,7 @@ from .inference import (
     RUST_WIDTH_RANK,
     is_rust_reference,
 )
+from .rust_ast import RustNode
 
 # allowed as names in Python but treated as keywords in Rust
 RUST_KEYWORDS = frozenset(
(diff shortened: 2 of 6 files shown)


adsharma:
The approach here allows gradual transition. Some visitors return strings (and will fail type check), the migrated ones will return a `RustNode`. When we're all done, we can remove the string -> ast -> string roundtrip in the visitor.

adsharma:
Already converted:

  - module/source file wrapper
  - simple calls
  - returns
  - assignments / let
  - augmented assignments
  - expression statements
  - functions
  - for, while, loop, normal if
  - asserts / macro calls
  - await / yield
  - print / drop

  Still mostly string-based:

  - expressions: binary ops, unary ops, comparisons, lambdas, attributes, subscripts, slices, tuples
  - literals/containers: lists, dicts, sets, bytes
  - structs/classes/enums/sealed classes
  - imports/usings/features/prologue
  - global lazy static collection initialization
  - generator/list comprehensions
  - try/unsupported exception output
  - extension module generation
  - dispatch/plugin functions still return strings
  - inherited py2many.clike visitors still build strings internally, then get wrapped as RustNode

  Also worth improving:

  - Replace the __getattribute__ visitor wrapper with explicit node returns once enough visitors are converted. It works, but it is transitional and adds overhead.
  - Add stricter tests that direct visitor calls return specific node classes, not just RustNode.
  - Add a small renderer test per construct to prevent accidental string-only regressions.
  - Eventually stop subclassing str; use proper AST classes and render only at the final pass. Right now RustNode(str) is a compatibility bridge.

PR #792, opened 2026-05-28
Update pymojo to emit mojo v0.26.2.0

jayvdb:
With this change, only dict.mojo still fails:

````
SKIPPED [1] test_cli.py:298: dict.mojo doesnt compile
````
https://github.com/jayvdb/py2many/actions/runs/26552107473/job/78216176810

### DIFF
--- a/pymojo/rewriters.py
+++ b/pymojo/rewriters.py
@@ -24,6 +24,34 @@ class MojoImplicitConstructor(ast.NodeTransformer):
         return node
 
 
+class MojoDictKeysInElider(ast.NodeTransformer):
+    """Rewrite `x in d.keys()` to `x in d`.
+
+    Mojo's `Dict.keys()` returns `_DictKeyIter`, which doesn't implement
+    `__contains__`. In Python both forms are equivalent for dicts, so dropping
+    the `.keys()` call is semantics-preserving and lets the emission go through
+    Mojo's `Dict.__contains__`.
+    """
+
+    def visit_Compare(self, node):
+        self.generic_visit(node)
+        if (
+            len(node.ops) == 1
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: tests/expected/bubble_sort.mojo [hunk 1]
```diff
@@ -1,7 +1,7 @@
-import testing
+from std.testing import assert_true
 
 
-fn bubble_sort(owned seq: List[Int]) -> List[Int]:
```
adsharma, line 4: The previous semantics were clear: `unsorted` was owned by `main()` and it transferred it to `bubble_sort()`, which returned a new array while transferring ownership back to main().

It's possible that the same thing is happening in the new code also, but its semantics aren't as clear as the previous code.

I haven't followed what's happening with mojo's borrow checker since the original announcement. Was this necessitated by some change on the mojo side?

adsharma: Looks like `owned` was 2024 and `var` is 2026. All good.

pymojo/rewriters.py
```diff
@@ -24,6 +24,34 @@ def visit_ClassDef(self, node):
...
+        if (
+            len(node.ops) == 1
+            and isinstance(node.ops[0], ast.In)
+            and len(node.comparators) == 1
+        ):
+            rhs = node.comparators[0]
+            if (
+                isinstance(rhs, ast.Call)
+                and isinstance(rhs.func, ast.Attribute)
+                and rhs.func.attr == "keys"
```
adsharma, line 47: assert that it was actually a `dict` and not some other duck type.

adsharma:
mojo updated to 1.0.0

PR #795, opened 2026-05-28
Add .mise.toml

jayvdb:
Allows installing all necessary tools using https://mise.jdx.dev/ , and run the tests.

Here it is running in GHA:
https://github.com/jayvdb/py2many/actions/runs/26556459722/job/78229615426

Here is the possible workflow we can use https://github.com/jayvdb/py2many/blob/mise-no-xdist-gha/.github/workflows/mise.yml

I have fixes for the macos and windows failures/incorrect skips.

17 mins to run all tests on Linux.

When I also add [pytest-xdist](https://github.com/pytest-dev/pytest-xdist) , and fix the many test concurrency bugs, it goes down to 12mins to run _all_ tests on Linux.
https://github.com/jayvdb/py2many/actions/runs/26507313773/job/78062943321

### DIFF
--- a/.mise.toml
+++ b/.mise.toml
@@ -1,0 +1,120 @@
+[tools]
+"asdf:mise-plugins/mise-nim" = { version = "2.2.10", os = ["linux", "macos"] }
+# asdf plugins are shell scripts that can't run on Windows; use the official
+# prebuilt Windows binary there instead.
+"http:nim" = { version = "2.2.10", url = "https://nim-lang.org/download/nim-{{ version }}_x64.zip", bin_path = "nim-{{ version }}/bin", os = ["windows"] }
+# conda's clang++ isn't usable standalone on Windows: its DLLs (libLLVM etc.)
+# live in the conda env's Library/bin, not on PATH via mise's shim, and Windows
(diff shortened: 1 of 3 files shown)


adsharma:
It's good to have more tools and choice. I'll check it out later today (trying to get a LadybugDB release going).

One of the reasons for switching to a system that has fewer dependencies is the recent instability around GHA infra and attacks on Canonical toolchain/infra. We couldn't pull gcc-13 on ubuntu-22.04 for several days straight. That lead to pulling all the frequently used tools to a docker image, so we pull less from infra maintained by anyone else.

With support for 9-10 languages, there is also the issue of many moving parts that break py2many for no reason other than the language made some incompatible changes. It's hard to keep up.

jayvdb:
All tests green on all platforms https://github.com/jayvdb/py2many/actions/runs/26621023936 , except for tests that are failing on main , listed at https://github.com/py2many/py2many/commit/c4de1855dc6dc889007aa86f32e012c9ba244552

adsharma (approved):
mise seems to be a good tool to try. Let's see how it goes.

PR #800, opened 2026-05-29
Pin dfmt version

jayvdb:
Workaround https://github.com/dlang/dub/issues/1113


jayvdb:
This increases the linux mise run time by 1 min, reduces macos from 24m to 19m, and windows from 54m to 32m.

But, most importantly, it is critical in order to use pytest-xdist.

PR #815, opened 2026-06-26
Add SMT features to Lean transpiler

jayvdb:


adsharma:
Suppose a python programmer makes a change and wants to verify if the pre/post conditions and the invariants still hold.

With SMT, the [flow](https://github.com/py2many/static-python-skill/blob/main/reference/verification.md) we have is:

```
uvx py2many --smt test.py - | z3 -smt2 -in
```

No counter example means you're good to go.

With Lean, the equivalent flow is:

```
lake build
```

If it compiles, the invariant is not broken. But for that we need a `lakefile.toml`. How about adding a `scripts/lean-runner.sh`?

adsharma:
There are also constructs like:

```
theorem deposit_balance (acct : BankAccount) (amt : Int) (h : amt > 0) :
    (acct.deposit amt h).balance = acct.balance + amt := rfl

theorem deposit_increases (acct : BankAccount) (amt : Int) (h : amt > 0) :
    (acct.deposit amt h).balance > acct.balance := by
  simp [BankAccount.deposit]; omega

-- order of two deposits doesn't affect the final balance
theorem deposit_order_irrelevant (acct : BankAccount) (a b : Int) (ha : a > 0) (hb : b > 0) :
    ((acct.deposit a ha).deposit b hb).balance =
    ((acct.deposit b hb).deposit a ha).balance := by
  simp [BankAccount.deposit]; omega
```

It's not clear how to represent them in an approachable way for a python programmer. We can come to that later.

PR #830, opened 2026-08-03
lean: generate real postconditions from CHECKER.post blocks

### LINKED ISSUE #829: lean: verified_sort.lean is missing a post condition
Use of the `result` keyword and suppressing via noqa is also bogus

adsharma:
Fixes: #829

### DIFF
--- a/py2many/rewriters.py
+++ b/py2many/rewriters.py
@@ -551,6 +551,128 @@ class LoopElseRewriter(ast.NodeTransformer):
         node.body = body
 
 
+class SelfMutatingMethodRewriter(ast.NodeTransformer):
+    """Rewrite methods that mutate ``self`` fields and ``return self`` into the
+    equivalent functional form ``return ClassName(...)``.
+
+    Backends like Lean and SMT have no in-place object
+    mutation, so ``self.balance += amount; return self`` must become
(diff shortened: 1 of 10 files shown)


py2many/rewriters.py
```diff
@@ -551,6 +551,128 @@ def _visit_Scope(self, node) -> Any:
...
+    ``self.a += x; self.b = self.a + 1; return self`` becomes
+    ``return C(self.a + x, self.a + 1)`` for fields ``(a, b)``.
+    """
+
+    def __init__(self, language: str):
+        super().__init__()
+        # Python keeps in-place mutation; Rust expresses it natively via a
+        # ``&mut self`` receiver.  The remaining backends get the functional
+        # rewrite.
+        self._enable = language not in ("python", "rust")
```
adsharma: My guess is most languages support mutable self. Perhaps we should make an exception list for the inverse. @jayvdb - anything of the top of your head?

jayvdb: Go requires a pointer to mutate the object.

We should add a mutation to `classes` test case - that would sort this out properly.

adsharma: going with lean and smt as exceptions

